PR #1, opened 2020-04-09
Create network modification server

FranckLecuyer:


.github/workflows/maven.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Download cassandra 4.0-alpha3
```
geofjamg, line 19: You don't need Cassandra for this service?

README.md
```diff
@@ -1 +1,8 @@
-# network-modification-server
\ No newline at end of file
+# network-modification-server
+
+[![Actions Status](https://github.com/gridsuite/network-modification-server/workflows/CI/badge.svg)](https://github.com/gridsuite/network-modification-server/actions)
+[![Coverage Status](https://sonarcloud.io/api/project_badges/measure?project=org.gridsuite%3Anetwork-modification-server&metric=coverage)](https://sonarcloud.io/component_measures?id=org.gridsuite%3Anetwork-modification-server&metric=coverage)
+[![MPL-2.0 License](https://img.shields.io/badge/license-MPL_2.0-blue.svg)](https://www.mozilla.org/en-US/MPL/2.0/)
+[![Join the community on Spectrum](https://withspectrum.github.io/badge/badge.svg)](https://spectrum.chat/powsybl)
```
geofjamg, line 6: To remove this the powsybl chat

pom.xml
```diff
@@ -0,0 +1,284 @@
...
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+-->
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.gridsuite</groupId>
```
geofjamg, line 15: Use powsybl parent pom

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -0,0 +1,47 @@
...
+ * @author Franck Lecuyer <franck.lecuyer at rte-france.com>
+ */
+@RestController
+@RequestMapping(value = "/" + NetworkModificationApi.API_VERSION + "/")
+@Api(tags = "network-modification-server")
+@ComponentScan(basePackageClasses = NetworkModificationService.class)
+public class NetworkModificationController {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NetworkModificationController.class);
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
geofjamg, line 32: You don't need to create you own mapper

.github/workflows/maven.yml
FranckLecuyer: done

README.md
FranckLecuyer: reference to powsybl chat removed

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
FranckLecuyer: ObjectMapper removed

pom.xml
FranckLecuyer: I tried to use the powsybl parent pom, but now i got an error with checkstyle on the NetworkModificationApplication class : it says that this class is a utility class, and so it should not have a default constructor. So i try to add a private default constructor, but now it says that the class should be final, but it is not accepted at execution ....

Is there a lombok annotation to avoid this error with checkstyle ?

geofjamg: There is a way to discard this checkstyle error with a comment. You can have a look at the other web services for that.

FranckLecuyer: It seems that the other web services currently don't use the powsybl parent pom, and in their own checkstyle.xml file, they disactivate the 'HideUtilityClassConstructor' check, but this check is present in the powsybl parent checkstyle xml file

So, should i keep a checkstyle xml file in the new modification-server service, or else do you know if there is a way to override the parent checkstyle configuration file, in order to suppress some checks ?

jonenst: Not all webservices have been migrated to the new parent pom :(
This one has been migrated: https://github.com/powsybl/powsybl-network-store/blob/6cebe249cb6d2d56224c3a30d2afa70ef22c1591/network-store-server/src/main/java/com/powsybl/network/store/server/NetworkStoreApplication.java#L6

FranckLecuyer: Done

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -0,0 +1,44 @@
...
+@Api(tags = "network-modification-server")
+@ComponentScan(basePackageClasses = NetworkModificationService.class)
+public class NetworkModificationController {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NetworkModificationController.class);
+
+    @Inject
+    private NetworkModificationService networkModificationService;
+
+    @PutMapping(value = "/switch/{networkUuid}/{switchId}")
```
geofjamg, line 34: To follow classical REST API convention it should be:
/networks/{networkUuid}/switches/{switchId}

FranckLecuyer: Modifications done

PR #17, opened 2021-02-19
Store elementary modifications in postgresql database

SlimaneAmar:
Signed-off-by: Slimane AMAR <slimane.amar@rte-france.com>


pom.xml
```diff
@@ -45,8 +45,10 @@
 
     <properties>
         <java.version>11</java.version>
-
         <gridsuite-dependencies.version>4</gridsuite-dependencies.version>
+        <postgres.version>13.1.0</postgres.version>
+        <zonky-embedded-database-test.version>2.0.0-beta1</zonky-embedded-database-test.version>
+        <embedded-postgres.version>1.2.10</embedded-postgres.version>
```
jonenst, line 51: why use embedded-postgres ? switch to h2 for tests ?

SlimaneAmar: When I tested several inheritance strategies with spring data
the generated schema was not compatible with the H2 dialect

src/main/java/org/gridsuite/modification/server/entities/ElementaryModificationEntity.java
```diff
@@ -0,0 +1,56 @@
...
+@Entity
+@Table(name = "elementaryModification")
+public class ElementaryModificationEntity extends AbstractModificationEntity {
+    @Column(name = "equipmentId")
+    private String equipmentId;
+
+    @Transient
+    private Set<String> substationIds = Set.of();
+
+    @OneToOne(fetch = FetchType.EAGER, optional = false, orphanRemoval = true, cascade = {CascadeType.ALL})
```
jonenst, line 33: we try to never use EAGER, can it be avoided ?

pom.xml
jonenst: is it compatible now with JOINs ? which ones were not compatible ? It's good to know probably worth adding in a comment

src/main/java/org/gridsuite/modification/server/DataSourceConfig.java
```diff
@@ -0,0 +1,38 @@
...
+ * @author Slimane Amar <slimane.amar at rte-france.com>
+ */
+@Configuration
+@PropertySource(value = {"classpath:database.properties"})
+@PropertySource(value = {"file:/config/database.properties"}, ignoreResourceNotFound = true)
+public class DataSourceConfig {
+
+    @Bean
+    public DataSource getDataSource(Environment env) {
+        String url = env.getRequiredProperty("scheme") + "://" + env.getRequiredProperty("hostPort")
```
jonenst, line 27: to be synchronized with the latest study-server version

pom.xml
SlimaneAmar: Yes now it is compatible
This is `INHERITS` which was not compatible
```
CREATE TABLE capitals (
    state           char(2)
) INHERITS (cities);
```

src/main/java/org/gridsuite/modification/server/entities/ElementaryModificationEntity.java
SlimaneAmar: Yes
I haven't watched yet the latest study-server

src/main/java/org/gridsuite/modification/server/DataSourceConfig.java
SlimaneAmar: Yes 
I 'm going to do it

pom.xml
SlimaneAmar: i 'm going to reuse h2

jonenst: What was the jpa configuration that made it generate inherits statements ?

src/main/java/org/gridsuite/modification/server/WebFluxConfig.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public void configureHttpMessageCodecs(ServerCodecConfigurer configurer) {
+        var objectMapper = objectMapper();
+        configurer.defaultCodecs().jackson2JsonEncoder(new Jackson2JsonEncoder(objectMapper));
+        configurer.defaultCodecs().jackson2JsonDecoder(new Jackson2JsonDecoder(objectMapper));
+    }
+
+    public static ObjectMapper createObjectMapper() {
+        var objectMapper = Jackson2ObjectMapperBuilder.json().build();
+        objectMapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
```
jonenst, line 34: why do we need this ? If I understand correctly, its to force jackson to serialize timestamps as datestrings instead of an integer. But we never have this in others microservices that deal with dates, what do we use instead? we should do the same everywhere (I think it's not very homogeneous already, between localdatetime and zoneddatetime... but adding timestamps is not going in the good direction I think)

SlimaneAmar: Removed

src/main/java/org/gridsuite/modification/server/DataSourceConfig.java
SlimaneAmar: Done

pom.xml
SlimaneAmar: H2 used  now

src/main/java/org/gridsuite/modification/server/entities/ElementaryModificationEntity.java
SlimaneAmar: Done

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -0,0 +1,123 @@
...
+    private ModificationGroupRepository modificationGroupRepository;
+
+    private ModificationRepository modificationRepository;
+
+    public NetworkModificationRepository(ModificationGroupRepository modificationGroupRepository, ModificationRepository modificationRepository) {
+        this.modificationGroupRepository = modificationGroupRepository;
+        this.modificationRepository = modificationRepository;
+    }
+
+    public void deleteAll() {
```
jonenst, line 39: transactional ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -0,0 +1,123 @@
...
+    public List<ElementaryModificationInfos> getElementaryModifications(UUID groupUuid) {
+        ModificationGroupEntity group = getModificationGroup(groupUuid);
+        return this.modificationRepository.findAllByGroupUuid(group.getUuid())
+                .stream()
+                .filter(m -> m.getType().equals(ModificationType.ELEMENTARY.name()))
+                .map(ElementaryModificationEntity.class::cast)
+                .map(ElementaryModificationEntity::toElementaryModificationInfos)
+                .collect(Collectors.toList());
+    }
+
```
jonenst, line 84: transactional ?

jonenst:
We are missing an index on modification.group_uuid
Also, how many modifications can be in a group ? If more than 100, then we need to avoid the n+1 query on findAllByGroupUuid, (on query per attribute)

src/main/java/org/gridsuite/modification/server/service/NetworkStoreListener.java
```diff
@@ -0,0 +1,86 @@
...
+            substationsIds.add(((ThreeWindingsTransformer) identifiable).getTerminal(ThreeWindingsTransformer.Side.TWO).getVoltageLevel().getSubstation().getId());
+            substationsIds.add(((ThreeWindingsTransformer) identifiable).getTerminal(ThreeWindingsTransformer.Side.THREE).getVoltageLevel().getSubstation().getId());
+        }
+
+        return substationsIds;
+    }
+
+    @Override
+    public void onUpdate(Identifiable identifiable, String attribute, Object oldValue, Object newValue) {
+        storeModification(identifiable, attribute, newValue);
```
jonenst, line 65: maybe also store the old value so that modifications can be undone ?

SlimaneAmar:
> We are missing an index on modification.group_uuid

As group uuid is a primary key is it necessary to explicitly create an index for it ?

src/main/java/org/gridsuite/modification/server/entities/ElementaryModificationEntity.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+@NoArgsConstructor
+@Getter
+@Entity
+@Table(name = "elementaryModification")
+public class ElementaryModificationEntity extends ModificationEntity {
+    @Column(name = "equipmentId")
+    private String equipmentId;
+
+    @Transient
```
jonenst, line 29: why Transient ? If we don't want to store it, but still return it to the http request creating the modification, then we should refactor the code. It's not simple to remember that some fields are valid only when the entity is created, not when it is loaded.
Or do we want to persist this information ?

src/main/java/org/gridsuite/modification/server/entities/ModificationEntity.java
```diff
@@ -0,0 +1,64 @@
...
+ * @author Slimane Amar <slimane.amar at rte-france.com>
+ */
+@NoArgsConstructor
+@Getter
+@Entity
+@Inheritance(strategy = InheritanceType.JOINED)
+@Table(name = "modification")
+public class ModificationEntity {
+
+    public static final String DATE_COLUMN_NAME = "date";
```
jonenst, line 31: why single out this column name ?

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.stereotype.Repository;
+
+/**
+ * @author Slimane Amar <slimane.amar at rte-france.com>
+ */
+@Repository
+public interface ModificationRepository extends JpaRepository<ModificationEntity, UUID> {
+    @Query(value = "SELECT *, 0 AS clazz_ FROM modification WHERE group_uuid = ?1", nativeQuery = true)
```
jonenst, line 22: can we use @DiscriminatorColumn and @DiscriminatorValue to avoid the magic of "0" and "clazz_" ?

src/main/java/org/gridsuite/modification/server/entities/ElementaryModificationEntity.java
SlimaneAmar: Exact
I'm going to return only these information for the http request

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -0,0 +1,26 @@
...
+    List<ModificationEntity> findAll(UUID uuid);
```
jonenst, line 23: rename to a more explicit name saying that it is not a polymorphic query (only the base data is retrieved). Do we really need this ? If we do a DTO projection on the base class fields only, does it still do the joins ?

src/main/java/org/gridsuite/modification/server/entities/ModificationEntity.java
SlimaneAmar: Used in a previous version to sort the modifications
I'm going to delete it

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
jonenst: or can we remove inheritance and replace it with just relations ? For exemple ElementaryModificationEntity would have only the data regarding the actual modification, and a one to one mapping to a Modification which has the generic data.

This would still allow to select the base attribute like in this query, but not allow to select both elementaryModifications values and other modification specific values.

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -0,0 +1,124 @@
...
+        this.modificationRepository = modificationRepository;
+    }
+
+    @Transactional // To have the 2 delete in the same transaction (atomic)
+    public void deleteAll() {
+        modificationRepository.deleteAll();
+        modificationGroupRepository.deleteAll();
+    }
+
+    @Transactional // To have the 2 create in the same transaction (atomic)
```
jonenst, line 45: don't forget that transactions also affect selects !

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
SlimaneAmar: Yes, we need to get general information about group modifications  like convergence
I tested with a base DTO class and JPA still do joins

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
SlimaneAmar: How ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -0,0 +1,124 @@
...
+                .collect(Collectors.toList());
+    }
+
+    @Transactional(readOnly = true)
+    // Due to the lazy loading done outside transaction (demarcation done in findAllByGroupUuid method)
+    public List<ElementaryModificationInfos> getElementaryModifications(UUID groupUuid) {
+        ModificationGroupEntity group = getModificationGroup(groupUuid);
+        return this.modificationRepository.findAllByGroupUuid(group.getUuid())
+                .stream()
+                .filter(m -> m.getType().equals(ModificationType.ELEMENTARY.name()))
```
jonenst, line 79: Do this filter in the sql ?

SlimaneAmar:
>Also, how many modifications can be in a group ? If more than 100, then we need to avoid the n+1 query on findAllByGroupUuid, (on query per attribute)

Yes i think in cvg a group have a lot of modifications.
Since we need to retrieve either 
- the general informations of the modifications of a group
- the detailed information of a modification
i  think that the entity model of the elementary modifications is not good.

As there is a bijection between a modification and an attribute the first version was optimized :
```
public class ElementaryModificationEntity extends ModificationEntity {
    private String equipmentId;
   
    private String attributeName;
    private String attributeValue;
    private String attributeType;
}
```

src/main/resources/network-modification.sql
```diff
@@ -0,0 +1,100 @@
...
+       foreign key (id) 
+       references attribute;
+
+    alter table if exists modification 
+       add constraint group_uuid_fk_constraint 
+       foreign key (group_uuid) 
+       references modificationGroup;
+
+    alter table if exists stringAttribute 
+       add constraint FKib9wv30mqmp4n9t71yiggcs5f
```
jonenst, line 98: missing constraint name in annotations?

src/main/resources/network-modification.sql
```diff
@@ -0,0 +1,100 @@
+create sequence hibernate_sequence start 1 increment 1;
+
+    create table attribute (
+       id int8 not null,
+        attributeName varchar(255),
```
jonenst, line 5: not null

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
SlimaneAmar: > rename to a more explicit name saying that it is not a polymorphic query 

findAllBases ?
findAllBasesByGroupUuid ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
SlimaneAmar: I think we will never need to get only elementary modifications but rather all modifications in a group

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
SlimaneAmar: Only for InheritanceType.SINGLE_TABLE ?

src/main/java/org/gridsuite/modification/server/entities/ElementaryModificationEntity.java
```diff
@@ -0,0 +1,59 @@
...
+    private Set<String> substationIds = Set.of();
+
+    @OneToOne(cascade = CascadeType.ALL, fetch = FetchType.LAZY, optional = false, orphanRemoval = true)
+    @JoinColumn(name = "attribute_id",
+            referencedColumnName = "id",
+            foreignKey = @ForeignKey(
+                    name = "attribute_id_fk_constraint"
+            ))
+    private AbstractAttributeEntity attribute;
```
jonenst, line 38: can we avoid having an ID for the attribute, and instead directly embed it in the elementarymodification ?

jonenst:
```
    create table modification (
       uuid uuid not null,
        date timestamp,
        type varchar(255),
        group_uuid uuid not null,
        primary key (uuid)
    );
```
group_uuid is not a PK. for foreign keys, you also need to create the index yes

SlimaneAmar:
>group_uuid is not a PK. for foreign keys, you also need to create the index yes

Done

src/main/java/org/gridsuite/modification/server/entities/ElementaryModificationEntity.java
SlimaneAmar: The JPA spec does not support embeddable objects having inheritance

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
jonenst: The semantics are defined by isolation levels, but for example for REPEATABLE_READS if you wrap your function in a transaction you will get the value at the begining of the function, whereas if you don't wrap in a transaction you will get the value "at the moment of the select". Many more complex scenarios are possible..
I was just saying that because the comment says "2 inserts", but there is also 1 select in this function

SlimaneAmar: Yes but is not this what we want to avoid the group being deleted by another transaction

src/main/java/org/gridsuite/modification/server/entities/ModificationEntity.java
SlimaneAmar: Done

src/main/resources/network-modification.sql
SlimaneAmar: Done

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
SlimaneAmar: Done

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
SlimaneAmar: Done (renaming)

src/main/java/org/gridsuite/modification/server/entities/ElementaryModificationEntity.java
SlimaneAmar: Done

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
```diff
@@ -13,43 +16,82 @@
...
+    }
+
+    public Flux<UUID> getModificationGroups() {
+        return Flux.fromIterable(modificationRepository.getModificationGroupsUuids());
+    }
 
-    private Set<String> doModification(Network network, Runnable modification) {
-        return doModification(network, modification, MODIFICATION_ERROR);
+    public Flux<ModificationInfos> getModifications(UUID groupUuid) {
+        return Flux.fromIterable(modificationRepository.getModifications(groupUuid));
```
jonenst, line 75: do all db access during subscribe, not during mono creation

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
```diff
@@ -13,43 +16,82 @@
...
     }
 
-    private Set<String> doModification(Network network, Runnable modification, NetworkModificationException.Type typeIfError) {
+    public Mono<ElementaryModificationInfos> getElementaryModification(UUID groupUuid, UUID modificationUuid) {
+        return Mono.just(modificationRepository.getElementaryModification(groupUuid, modificationUuid));
```
jonenst, line 79: same, do all db access during subscribe, not during mono creation

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
SlimaneAmar: Done

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
SlimaneAmar: Done

jonenst (approved):

PR #39, opened 2021-10-07
Delete equipment

FranckLecuyer:


src/main/java/org/gridsuite/modification/server/elasticsearch/EquipmentInfosServiceImpl.java
```diff
@@ -31,7 +34,11 @@ public EquipmentInfos add(@NonNull EquipmentInfos equipmentInfos) {
 
     @Override
     public void delete(@NonNull String equipmentId, @NonNull UUID networkUuid) {
-        equipmentInfosRepository.deleteByEquipmentIdAndNetworkUuid(equipmentId, networkUuid);
+        try {
+            equipmentInfosRepository.deleteByEquipmentIdAndNetworkUuid(equipmentId, networkUuid);
+        } catch (Exception e) {
```
geofjamg, line 39: Why could be the reason for catching this exception. The problem is that in case of ES issue on the platform (service down), we won't any feeback (nobody look at error logs), the data won't be deleted ever. Why not doing the same thing with Cassandra or PostGres?

geofjamg (changes requested):

jonenst (approved):

PR #65, opened 2021-12-23
Fix : run the  embedded elasticsearch only for the unit tests that need it

SlimaneAmar:


src/test/java/org/gridsuite/modification/server/EmbeddedElasticsearch.java
```diff
@@ -24,9 +25,12 @@
 
     private static ElasticsearchContainer elasticsearchContainer;
 
+    @Value("${spring.data.elasticsearch.enabled:false}")
+    boolean elasticsearchEnabled;
+
     @PostConstruct
     public void postConstruct() {
-        if (elasticsearchContainer != null) {
+        if (elasticsearchContainer != null || !elasticsearchEnabled) {
```
niconoir, line 33: What if `!elasticsearchEnabled` **AND** `elasticsearchContainer != null` ? Maybe it should be ensured the elasticsearchContainer is stopped in this case ?

SlimaneAmar: Only one embedded elasticsearch is started for all test classes (static attribute)
And all EmbeddedElasticsearch beans are destroyed in the end of all test classes

niconoir (approved):

PR #69, opened 2022-01-11
Implement delete network modification

JacquesBorsenberger:


src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -72,6 +72,13 @@ public NetworkModificationController(NetworkModificationService networkModificat
         return ResponseEntity.ok().body(networkModificationService.getModifications(groupUuid, true));
     }
 
+    @DeleteMapping(value = "/groups/{groupUuid}/modifications/{modificationUuid}")
+    @Operation(summary = "Get modifications list of a group")
+    @ApiResponse(responseCode = "200", description = "Modification deleted")
+    public ResponseEntity<Mono<Void>> deleteModifications(@Parameter(description = "Group UUID") @PathVariable("groupUuid") UUID groupUuid, @PathVariable("modificationUuid") UUID modificationUuid) {
```
sBouzols, line 78: should we add this for the second parameter ?:
`@Parameter(description = "Modification UUID")`

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -72,6 +72,13 @@ public NetworkModificationController(NetworkModificationService networkModificat
...
+        return ResponseEntity.ok().body(networkModificationService.deleteModification(groupUuid, modificationUuid));
+    }
+
```
sBouzols, line 81: should we add an API to delete multiple modifications ? because `modificationRepository` allows it ...

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
```diff
@@ -270,6 +270,28 @@ public void testSwitch() {
         switchModifications(uriString + "&variantId=" + NetworkCreation.VARIANT_ID, "break1Variant", "notFound", "disc1Variant", "break2Variant", Set.of("s1Variant"), Set.of("s2Variant"), 8);
     }
 
+    @Test
+    public void testDeleteModification() {
```
sBouzols, line 274: Maybe add some test cases :
* to delete an non-existing modification
* to delete an existing modification but from an non-existing group 
What about error managment ?

sBouzols (changes requested):
some comments

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
JacquesBorsenberger: will add it when we need it

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -72,6 +72,13 @@ public NetworkModificationController(NetworkModificationService networkModificat
         return ResponseEntity.ok().body(networkModificationService.getModifications(groupUuid, true));
     }
 
+    @DeleteMapping(value = "/groups/{groupUuid}/modifications/{modificationUuid}")
+    @Operation(summary = "Get modifications list of a group")
```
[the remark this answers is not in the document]
JacquesBorsenberger: done

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
JacquesBorsenberger: done

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
JacquesBorsenberger: done

niconoir (approved):
Not tested

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
sBouzols: ok

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -184,12 +184,14 @@ public void deleteModificationGroup(UUID groupUuid) {
     }
 
     @Transactional // To have the find and delete in the same transaction (atomic)
-    public void deleteModifications(UUID groupUuid, Set<UUID> uuids) {
+    public int deleteModifications(UUID groupUuid, Set<UUID> uuids) {
         List<ModificationEntity> modifications = this.modificationRepository.findAllByGroupId(groupUuid)
                 .stream()
                 .filter(m -> uuids.contains(m.getId()))
                 .collect(Collectors.toList());
+        int count = modifications.size();
```
sBouzols, line 192: Is this really safe ? I mean deleteAll could fails, (it throws IllegalArgumentException by the way if entities is null) but this still return a desync count of elements (the result of findAllByGroupId)... Maybe it's more robust tu use a Try/Catch here and return only modifications.size() if the deleteAll succeed ?

sBouzols (changes requested):
a robustness change ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
JacquesBorsenberger: if deleteAll throws, we don't return anything, the exception is propagated.

sBouzols: Ha you right, my bad ! It's still seems strange to me to return the count returned before the deleteAll call but ok then.

sBouzols (approved):
Code Review ok
will test with all branches

PR #80, opened 2022-03-07
Fix search pb after deleting an equipment

FranckLecuyer:


niconoir (approved):
Code review and tests ok

PR #86, opened 2022-04-07
[WIP] Addition of network modification "load modification"

niconoir:
Creation of a network modification giving the possibility to modify loads attributes.


src/main/java/org/gridsuite/modification/server/entities/equipment/modification/BasicEquipmentModificationEntity.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+@NoArgsConstructor
+@Getter
+@MappedSuperclass
+public class BasicEquipmentModificationEntity extends EquipmentModificationEntity {
+    @Column(name = "equipmentNameValue")
+    private String equipmentNameValue;
+
+    @Column(name = "equipmentNameOp")
+    private OperationType equipmentNameOp;
```
FranckLecuyer, line 29: Maybe store OperationType enum as string in the database, instead of an int, with the following annotation : 
@Enumerated(EnumType.STRING)

Like we did in NodeEntity for example

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
```diff
@@ -643,6 +643,77 @@ private Mono<Void> assertLoadCreationInfosNotEmpty(LoadCreationInfos loadCreatio
...
+                    subReporter.report(Report.builder()
+                            .withKey("loadModification")
+                            .withDefaultMessage(exc.getMessage())
+                            .withValue("id", loadModificationInfos.getEquipmentId())
+                            .withSeverity(new TypedValue("LOAD_MODIFICATION_ERROR", TypedValue.ERROR_LOGLEVEL))
+                            .build());
+                }
+            }
+
+            // add the load creation entity to the listener
```
FranckLecuyer, line 695: Typo : add the load modification entity to the listener

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/modification/server/entities/equipment/modification/BasicEquipmentModificationEntity.java
niconoir: All OperationType enums are now stored as Strings.
Did not store LoadType as String for now, to be consistent with the way it is currently stored in LoadCreationEntity.

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
niconoir: Corrected

jonenst (approved):

FranckLecuyer (approved):

PR #99, opened 2022-06-24
New spring cloud stream channel for failed node builds

archendel:
Signed-off-by: LE SAULNIER Kevin <kevin.lesaulnier@rte-france.com>

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/service/BuildWorkerService.java
+++ b/src/main/java/org/gridsuite/modification/server/service/BuildWorkerService.java
@@ -36,7 +36,7 @@ import java.util.function.Consumer;
 import java.util.logging.Level;
 
 import static org.gridsuite.modification.server.service.BuildStoppedPublisherService.CANCEL_MESSAGE;
-import static org.gridsuite.modification.server.service.BuildStoppedPublisherService.FAIL_MESSAGE;
+import static org.gridsuite.modification.server.service.BuildFailedPublisherService.FAIL_MESSAGE;
 
 /**
  * @author Franck Lecuyer <franck.lecuyer at rte-france.com>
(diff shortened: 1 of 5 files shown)


src/main/java/org/gridsuite/modification/server/service/BuildWorkerService.java
```diff
@@ -175,6 +181,32 @@ public Consumer<Flux<Message<String>>> consumeCancelBuild() {
             .subscribe();
     }
 
+    @Bean
+    public Consumer<Flux<Message<String>>> consumeFailedBuild() {
```
SlimaneAmar, line 185: Why ?

SlimaneAmar (changes requested):

SlimaneAmar (approved):

src/main/java/org/gridsuite/modification/server/service/BuildWorkerService.java
```diff
@@ -58,21 +58,27 @@ public class BuildWorkerService {
 
     private BuildStoppedPublisherService stoppedPublisherService;
 
+    private BuildFailedPublisherService failedPublisherService;
+
     private Map<String, CompletableFuture<List<ModificationInfos>>> futures = new ConcurrentHashMap<>();
 
     private Map<String, BuildCancelContext> cancelBuildRequests = new ConcurrentHashMap<>();
 
+    private Map<String, BuildFailedContext> failedBuildRequests = new ConcurrentHashMap<>();
```
YenguiSeddik, line 67: I think this map is not used any more. maybe it should be deleted

archendel: Fixed

src/main/java/org/gridsuite/modification/server/service/BuildWorkerService.java
archendel: Fixed -> removed

YenguiSeddik (approved):

PR #165, opened 2022-12-05
add substation properties from front

laurgarn:


src/main/java/org/gridsuite/modification/server/entities/equipment/creation/SubstationCreationEntity.java
```diff
@@ -27,11 +30,18 @@ public class SubstationCreationEntity extends EquipmentCreationEntity {
     @Column(name = "country")
     private Country country;
 
-    public SubstationCreationEntity(String equipmentId, String equipmentName, Country country) {
+    @ElementCollection(fetch = FetchType.EAGER)
```
FranckLecuyer, line 33: why EAGER ?

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/modification/server/entities/equipment/creation/SubstationCreationEntity.java
laurgarn: Because it forgot them when I was on lazy mode.

FranckLecuyer (approved):

ayolab (approved):

PR #169, opened 2022-12-14
add incremental node build when we append some modifications (copy or cut and paste)

dbraquart:


FranckLecuyer (approved):
Review: OK
Tests: OK

FranckLecuyer (changes requested):
Tests: NOK (see remark in study-server PR :  https://github.com/gridsuite/study-server/pull/313)

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -63,16 +62,24 @@ public ResponseEntity<Void> createModificationGroup(@RequestParam("groupUuid") U
...
+                                                              @RequestBody List<UUID> modificationsUuidList) {
         switch (action) {
             case COPY:
-                return ResponseEntity.ok().body(networkModificationService.duplicateModifications(groupUuid, modificationsUuidList));
+                return ResponseEntity.ok().body(networkModificationService.duplicateModifications(groupUuid, networkUuid, reportUuid, reporterId, variantId, modificationsUuidList));
             case MOVE:
-                networkModificationService.moveModifications(groupUuid, originGroupUuid, before, modificationsUuidList);
+                // 2 params are optional
+                UUID sourceGroupUuid = originGroupUuid == null ? groupUuid : originGroupUuid;
+                boolean canBuildNode = build == null || build;
```
archendel, line 81: it means the default behaviour is to build the node
I think it is better to add default value to the param like this :
**@RequestParam(value = "build", required = false, defaultValue = "true") Boolean build**

archendel, line 81: Does it mean we can move a modification without changing it's group uuid, and still be in incremental build ?

In the below case, I think it could cause bug :
m1 -> m2 -> m3

We move m1 after m3 with build = true, we would apply m1 again on this node ?

archendel (changes requested):

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
dbraquart: yes, I will use defaultValue here

dbraquart: Yes we could do such wrong case (cut/paste in same node, with rebuild).
But we should not call this rest service directly: only study-server does, with all the logic.
Should I add some logic here ?
- I could add : no possible build when we are in the same node (easy)
- I can't add : no possible build when the target node belongs to the source node tree (only study-server know that)

FranckLecuyer (approved):

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
archendel: It's a good question, I believe each service should be independent and not rely on other ones controls
But I might be wrong, could you check with a tech lead ?

I validate the PR for now

archendel (approved):

PR #180, opened 2022-12-28
WIP Add some fields related to the startup for the GeneratorCreationEntity

flomillot:
Signed-off-by: Florent MILLOT <millotflo@gmail.com>
#1474


src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
```diff
@@ -2065,16 +2134,7 @@ public void testDuplicateModificationGroup() throws Exception {
                 post(URI_NETWORK_MODIF_BUS_BREAKER)
                     .content(objectWriter.writeValueAsString(vl1))
                     .contentType(MediaType.APPLICATION_JSON))
-            .andExpect(status().isOk())
-            .andReturn().getResponse().getContentAsString();
-
-        VoltageLevelCreationInfos vl2 = ModificationCreation.getCreationVoltageLevel("s1", "vl2Id", "vl2Name");
```
flomillot, line 2071: @SlimaneAmar for me it doesnt change anything, confirm me if I'm right or not

Meklo: Why was this request removed ?

flomillot: Because now we don't recreate a new network at each mock, and we can't insert 2 equipments with the same ID.
So I removed it because it was causing this error and for me it wasn't really important for the test anymore.

Meklo: Fair enough

Meklo (approved):

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
flomillot: This thread is about ModificationControllerTest, you mean in GeneratorCreation ?
Yeah it need to be refactorized.

flomillot: Why don't we just call addExtensionsToGenerator inside createGeneratorInBusBreaker ?

flomillot: Done

PR #182, opened 2022-12-29
migration to new design substation

anistouri:
Signed-off-by: TOURI ANIS <anis-1.touri@rte-france.com>


src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -300,7 +298,13 @@ public EquipmentCreationEntity createTwoWindingsTransformerEntity(String id, Str
     }
 
     public EquipmentCreationEntity createSubstationEntity(String id, String name, Country country, Map<String, String> properties) {
-        return new SubstationCreationEntity(id, name, country, properties);
+        return new SubstationCreationEntity(SubstationCreationInfos.builder()
```
laurgarn, line 301: You can remove the 'new'

src/main/java/org/gridsuite/modification/server/entities/equipment/creation/SubstationCreationEntity.java
```diff
@@ -63,5 +70,6 @@ public SubstationCreationInfos toSubstationCreationInfos() {
                 .equipmentName(getEquipmentName())
                 .substationCountry(getCountry()).properties(getProperties());
     }
+
```
laurgarn, line 73: It lacks a
```
@Override
public void cloneWithIdsToNull() {
    super.cloneWithIdsToNull();
    if (this.properties != null) {
        this.properties = new HashMap<>(properties);
    }
}
```
for (forgotten) test for properties in matcher to succeed

laurgarn: An another mean, as it seems cloneWithIdsToNull is part of the legacy interface we do not want to use anymore.

laurgarn: I added it myself. The trick around JPA, with this cloneWithIdsToNull will have to be re-evaluated.

laurgarn (approved):

SlimaneAmar (approved):

PR #211, opened 2023-02-15
Fix lazy loading in transaction when we move some modifications

SlimaneAmar:


dbraquart (commented):
I cannot see the problem on the main branch, so its hard for me to understand the fix
EDIT: I can see it with a "Create Group" modification

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -51,8 +52,7 @@ public void saveModifications(UUID groupUuid, List<? extends ModificationEntity>
     }
 
     @Transactional
-    // To have all move in the same transaction (atomic)
-    // When we move modifications, we move them right before referenceModification when it is defined, at the end of list otherwise
+    // TODO Remove transaction when errors will no longer be sent to the front
     public List<ModificationEntity> moveModifications(UUID destinationGroupUuid, UUID originGroupUuid, List<UUID> modificationsUuid, UUID referenceModificationUuid) {
```
dbraquart, line 56: the caller (NetworkModificationRepository::moveModifications) is now Transactional, so why not removing Transactional here now ?

dbraquart (changes requested):
tests: OK
code: just a question about Transactional flag

dbraquart (approved):

PR #221, opened 2023-03-23
When creating 'BusbarSectionPosition' extension the positions are now checked

SlimaneAmar:


dbraquart (approved):
code ok
tests not done

PR #229, opened 2023-04-14
Check that all requests mocked with WireMock are verified at the end of each test.

AAJELLAL:


src/test/java/org/gridsuite/modification/server/utils/WireMockUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.junit.Assert.assertEquals;
+
+public class WireMockUtils {
+
+    private final WireMockServer wireMock;
```
SlimaneAmar, line 24: Rename by wireMockServer

AAJELLAL: Done

FranckLecuyer (approved):

PR #243, opened 2023-05-09
Upgrade to GS dependencies 24

etiennehomer:


pom.xml
```diff
@@ -45,7 +45,7 @@
 
     <properties>
         <java.version>11</java.version>
-        <gridsuite-dependencies.version>23</gridsuite-dependencies.version>
+        <gridsuite-dependencies.version>24</gridsuite-dependencies.version>
```
SlimaneAmar, line 48: We shouldn't only just overload

SlimaneAmar (approved):

PR #253, opened 2023-05-23
Remove spring context parameter

SlimaneAmar:


jonenst (changes requested):
getting a spring bean from a static field is bad practice

jonenst (approved):

PR #263, opened 2023-06-12
add shunt compensator modification

YenguiSeddik:


src/main/java/org/gridsuite/modification/server/entities/equipment/modification/ShuntCompensatorModificationEntity.java
```diff
@@ -0,0 +1,92 @@
...
+    public ShuntCompensatorModificationInfos toModificationInfos() {
+        return toShuntCompensatorModificationInfosBuilder().build();
+    }
+
+    private ShuntCompensatorModificationInfos.ShuntCompensatorModificationInfosBuilder<?, ?> toShuntCompensatorModificationInfosBuilder() {
+        return ShuntCompensatorModificationInfos
+                .builder()
+                .uuid(getId())
+                .date(getDate())
+                .voltageLevelId(toAttributeModification(getVoltageLevelIdValue(), getVoltageLevelIdOp()))
```
dbraquart, line 85: we need voltageLevelId, but it's strange to get it with a modifiable field (value/op couple) because we can not change it with the form (it is not displayed either).

src/main/java/org/gridsuite/modification/server/modifications/ShuntCompensatorModification.java
```diff
@@ -0,0 +1,88 @@
...
+    private void applyModificationOnLinearModel(Reporter subReporter, ShuntCompensator shuntCompensator, VoltageLevel voltageLevel) {
+        List<Report> reports = new ArrayList<>();
+        ShuntCompensatorLinearModel model = shuntCompensator.getModel(ShuntCompensatorLinearModel.class);
+        var shuntCompensatorType = model.getBPerSection() > 0 ? ShuntCompensatorType.CAPACITOR : ShuntCompensatorType.REACTOR;
+
+        if (modificationInfos.getShuntCompensatorType() != null) {
+            reports.add(ModificationUtils.getInstance().buildModificationReport(shuntCompensatorType, modificationInfos.getShuntCompensatorType().getValue(), "Type"));
+            shuntCompensatorType = modificationInfos.getShuntCompensatorType().getValue();
+            if (modificationInfos.getQAtNominalV() == null) {
+                // we retrieve the absolute value of susceptance per section, then we determine the sign using the type
```
dbraquart, line 71: this IF is marked as not covered by Sonar. Could you add a test for it ?

dbraquart (changes requested):
code: 
- more testing needed I think
- voltageLevelId is stored as an AttributeModification while it should not

tests: the back-end and database are working fine

src/main/java/org/gridsuite/modification/server/dto/ShuntCompensatorModificationInfos.java
```diff
@@ -0,0 +1,69 @@
...
+@Schema(description = "Shunt compensator modification")
+public class ShuntCompensatorModificationInfos extends BasicEquipmentModificationInfos {
+
+    @Schema(description = "voltage level id")
+    private String voltageLevelId;
+
+    @Schema(description = "Susceptance per section")
+    private AttributeModification<Double> susceptancePerSection;
+
+    @JsonProperty("qAtNominalV")
```
sBouzols, line 43: Why do you need this @JsonProperty annotation here ?

src/main/java/org/gridsuite/modification/server/modifications/ShuntCompensatorModification.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public void apply(Network network, Reporter subReporter) {
+        ShuntCompensator shuntCompensator = network.getShuntCompensator(modificationInfos.getEquipmentId());
+        if (shuntCompensator == null) {
+            throw new NetworkModificationException(SHUNT_COMPENSATOR_NOT_FOUND,
+                    String.format("Shunt compensator %s does not exist in network", modificationInfos.getEquipmentId()));
+        }
+        VoltageLevel voltageLevel = network.getVoltageLevel(modificationInfos.getVoltageLevelId());
+
```
sBouzols, line 46: We don't throw an error if the VoltageLevel does not exist in the network ?

src/main/java/org/gridsuite/modification/server/modifications/ShuntCompensatorModification.java
```diff
@@ -0,0 +1,88 @@
...
+        var shuntCompensatorType = model.getBPerSection() > 0 ? ShuntCompensatorType.CAPACITOR : ShuntCompensatorType.REACTOR;
+
+        if (modificationInfos.getShuntCompensatorType() != null) {
+            reports.add(ModificationUtils.getInstance().buildModificationReport(shuntCompensatorType, modificationInfos.getShuntCompensatorType().getValue(), "Type"));
+            shuntCompensatorType = modificationInfos.getShuntCompensatorType().getValue();
+            if (modificationInfos.getQAtNominalV() == null) {
+                // we retrieve the absolute value of susceptance per section, then we determine the sign using the type
+                double bPerSectionAbsoluteValue = Math.abs(model.getBPerSection());
+                double newBPerSection = shuntCompensatorType == ShuntCompensatorType.CAPACITOR ? bPerSectionAbsoluteValue : -bPerSectionAbsoluteValue;
+                model.setBPerSection(newBPerSection);
```
sBouzols, line 74: Maybe we should add a report here as well ?

src/main/java/org/gridsuite/modification/server/dto/ShuntCompensatorModificationInfos.java
YenguiSeddik: Because the field began with lowercase followed by uppercase, the generated getter and setter are : setqAtNominalV and getqAtNominalV instead of setQAtNominalV and getQAtNominalV.therefore, the value of qAtNominalV will not be affected during deserialization. @JsonProperty indicate which prop to look for in th json

src/main/java/org/gridsuite/modification/server/modifications/ShuntCompensatorModification.java
YenguiSeddik: This is the case when the user change only the shunt compensator type. There is report for that above at line 68: 
reports.add(ModificationUtils.getInstance().buildModificationReport(shuntCompensatorType, modificationInfos.getShuntCompensatorType().getValue(), "Type"));

src/main/java/org/gridsuite/modification/server/modifications/ShuntCompensatorModification.java
YenguiSeddik: done

src/main/java/org/gridsuite/modification/server/modifications/ShuntCompensatorModification.java
dbraquart: this IF is still uncovered. Would be nice to cover.

dbraquart (changes requested):
code: ok, but a request to cover some code in apply()
tests: ok

dbraquart (approved):
code: ok
tests: not done with last commit

src/main/java/org/gridsuite/modification/server/modifications/ShuntCompensatorModification.java
sBouzols: Then we don't want to log that bPerSectionAbsoluteValue change its value from positive to negative ? Thats a functionnal question, maybe we should ask a PO ?

src/main/java/org/gridsuite/modification/server/dto/ShuntCompensatorModificationInfos.java
sBouzols: ok

sBouzols (approved):
Code Review OK
Tests OK

PR #286, opened 2023-07-06
Update section count in shunt compensator creation

Walid-Sahnoun:


.attach_pid1907792
FranckLecuyer, line 1: Remove this file

src/main/java/org/gridsuite/modification/server/entities/equipment/creation/ShuntCompensatorCreationEntity.java
```diff
@@ -56,12 +56,11 @@ public class ShuntCompensatorCreationEntity extends InjectionCreationEntity {
 
     public ShuntCompensatorCreationEntity(ShuntCompensatorCreationInfos creationInfos) {
         super(creationInfos);
-        maximumNumberOfSections = creationInfos.getMaximumNumberOfSections();
-        currentNumberOfSections = creationInfos.getCurrentNumberOfSections();
+        maximumNumberOfSections = 1;
```
FranckLecuyer, line 59: why removing currentNumberOfSections, if we decide to keep maximumNumberOfSections and isIdenticalSections ?

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/modification/server/entities/equipment/creation/ShuntCompensatorCreationEntity.java
Walid-Sahnoun: well i removed only from constructor because it's produce en error because we removed the attribute from JSON at the front side,

FranckLecuyer: So, why not adding a check like : 
currentNumberOfSections = creationInfos.getCurrentNumberOfSections() != null ? creationInfos.getCurrentNumberOfSections() : 1;

Same pb in update method : 
converting the null value of Integer currentNumberOfSections in the dto to an int value in the entity ...
converting the null value of Boolean isIdenticalSections in the dto to a boolean value in the entity ...

FranckLecuyer (changes requested):
See remarks

ayolab (approved):

PR #312, opened 2023-08-16
Do not throw exception anymore when all filters have bad IDs.

flomillot:
But only when all equipements of all filters are wrong.


EstherDarkish (approved):

Walid-Sahnoun (approved):

EstherDarkish (approved):
Code OK, tests OK

PR #320, opened 2023-09-06
Copy node and duplicate not stashed modifications

souissimai:


ayolab (approved):

etiennehomer:
You should add a stashed modification to testDuplicateModificationGroup(). So that this behavior is tested

etiennehomer (approved):

PR #335, opened 2023-09-25
Fix error when active power is null in Vsc creation

YenguiSeddik:


src/main/java/org/gridsuite/modification/server/modifications/VscCreation.java
```diff
@@ -78,18 +79,21 @@ public void apply(Network network, Reporter subReporter) {
...
                 .setR(modificationInfos.getDcResistance())
                 .setMaxP(modificationInfos.getMaximumActivePower())
-                .setActivePowerSetpoint(modificationInfos.getActivePower())
                 .setConvertersMode(modificationInfos.getConvertersMode())
                 .setConverterStationId1(converterStation1 != null ? converterStation1.getId() : null)
-                .setConverterStationId2(converterStation2 != null ? converterStation2.getId() : null)
-                .add();
+                .setConverterStationId2(converterStation2 != null ? converterStation2.getId() : null);
+
+        if (modificationInfos.getActivePower() != null) {
```
FranckLecuyer, line 92: activePowerSetpoint is always mandatory to create an hvdc ...

FranckLecuyer (changes requested):

PR #342, opened 2023-10-10
minor change on report message

AbdelHedhili:


AbdelHedhili:
Disclaimer: coverage is at 0% because I only changed a reporter label, and we don't do checks for those in the UTs.

Meklo (approved):

FranckLecuyer (approved):

PR #353, opened 2023-10-23
gridsuite dependencies 27

AbdelHedhili:


pom.xml
```diff
@@ -144,14 +143,10 @@
         <dependency>
             <groupId>com.powsybl</groupId>
             <artifactId>powsybl-network-store-client</artifactId>
-            <!-- FIXME: to be removed at next powsybl-ws-dependencies upgrade  -->
-            <version>${powsybl-network-store.version}</version>
         </dependency>
         <dependency>
```
etiennehomer, line 147: You can remove the whole dependency I think. This is used when we need to override the network-store-version. 
But we don't need it anymore when we're based directly on the gridsuite-dependencies versions

AbdelHedhili: No I can't. It is used in the repo and if I don't put it in the dependency section, it doesn't matter if it is in gridsuite-dependencies it won't be used.

AbdelHedhili: But the remarke is true for another dependency network-store-iidm-impl (I removed it)

antoinebhs (commented):
I think we should also remove the class RemoveSubstation as specified in the spec

AbdelHedhili:
> I think we should also remove the class RemoveSubstation as specified in the spec

I cannot since the constructor of the RemoveSubstation class in powsybl-core is package private, so we need to make it public first before we can remove this hack

etiennehomer:
> > I think we should also remove the class RemoveSubstation as specified in the spec
> 
> I cannot since the constructor of the RemoveSubstation class in powsybl-core is package private, so we need to make it public first before we can remove this hack

You can use the builder nope ?

AbdelHedhili:
> > > I think we should also remove the class RemoveSubstation as specified in the spec
> > 
> > 
> > I cannot since the constructor of the RemoveSubstation class in powsybl-core is package private, so we need to make it public first before we can remove this hack
> 
> You can use the builder nope ?

What builder ? There is none (or I don't see one at least)

AbdelHedhili:
> @AbdelHedhili https://github.com/powsybl/powsybl-core/blob/8d6cb73c57bf2d3095cea2c8873e014896462c52/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveSubstationBuilder.java ?

Oh... totally missed it, my bad.

src/main/java/org/gridsuite/modification/server/modifications/EquipmentDeletion.java
```diff
@@ -42,11 +42,13 @@ public void apply(Network network, Reporter subReporter) {
         if (identifiable instanceof Connectable) {
             new RemoveFeederBay(modificationInfos.getEquipmentId()).apply(network, true, subReporter);
         } else if (identifiable instanceof HvdcLine) {
+
```
etiennehomer, line 45: remove empty row

antoinebhs (approved):

PR #359, opened 2023-10-30
Add by formula modification

YenguiSeddik:


src/main/java/org/gridsuite/modification/server/modifications/ByFormulaModification.java
```diff
@@ -0,0 +1,144 @@
...
+                case MULTIPLICATION -> value1 * value2;
+                case DIVISION -> {
+                    if (value2 == 0) {
+                        throw new NetworkModificationException(NetworkModificationException.Type.BY_FORMULA_MODIFICATION_ERROR,
+                                "there is a division by 0 in one formula");
+                    } else {
+                        yield value1 / value2;
+                    }
+                }
+                case MODULUS -> value1 % value2;
```
EstherDarkish, line 140: This should not be a modulus operation, but a percentage

EstherDarkish (changes requested):

src/main/java/org/gridsuite/modification/server/modifications/ByFormulaModification.java
```diff
@@ -0,0 +1,187 @@
...
+            formulaReports.forEach(formulaSubReporter::report);
+        }
+    }
+
+    private void applyFormulaOnFilterEquipments(Network network,
+                                                Map<UUID, FilterEquipments> exportFilters,
+                                                List<Report> formulaReports,
+                                                FormulaInfos formulaInfos,
+                                                FilterInfos filterInfos) {
+        var filterEquipments = exportFilters.get(filterInfos.getId());
```
anistouri, line 91: To enhance code readability and maintainability, I suggest specifying the explicit return type instead of using var 
```suggestion
         FilterEquipments filterEquipments = exportFilters.get(filterInfos.getId());
```

src/main/java/org/gridsuite/modification/server/modifications/ByFormulaModification.java
```diff
@@ -0,0 +1,187 @@
...
+                .build());
+
+        formulaReports.add(Report.builder()
+                .withKey("numberOfValidEquipment" + formulaReports.size())
+                .withDefaultMessage(String.format("      Number of equipment modified : %s", filterEquipments.getIdentifiableAttributes().size()))
+                .withSeverity(TypedValue.INFO_SEVERITY)
+                .build());
+
+        if (!CollectionUtils.isEmpty(filterEquipments.getNotFoundEquipments())) {
+            var equipmentIds = String.join(", ", filterEquipments.getNotFoundEquipments());
```
anistouri, line 107:
```suggestion
           String equipmentIds = String.join(", ", filterEquipments.getNotFoundEquipments());
```

src/main/java/org/gridsuite/modification/server/modifications/ModificationUtils.java
```diff
@@ -974,5 +975,27 @@ private void createReactiveCapabilityCurvePoint(ReactiveCapabilityCurveAdder add
         addToReports(reports, point.getQminP(), "QminP" + fieldSuffix);
         addToReports(reports, point.getQmaxP(), "QmaxP" + fieldSuffix);
     }
+
+    public boolean isValidFilter(Reporter subReporter,
+                                 NetworkModificationException.Type errorType,
+                                 Map<UUID, String> filters,
```
anistouri, line 981: it's not used

src/main/java/org/gridsuite/modification/server/modifications/ModificationUtils.java
```diff
@@ -974,5 +975,27 @@ private void createReactiveCapabilityCurvePoint(ReactiveCapabilityCurveAdder add
...
+                                 Map<UUID, FilterEquipments> exportFilters) {
+        // collect all filters with wrong equipments ids
+        Map<UUID, FilterEquipments> filterWithWrongEquipmentsIds = exportFilters.entrySet()
```
anistouri, line 984: this variable is not used

src/main/java/org/gridsuite/modification/server/service/FilterService.java
```diff
@@ -65,6 +70,15 @@ public List<FilterEquipments> exportFilters(List<UUID> filtersUuids, UUID networ
         }
     }
 
+    public Map<UUID, FilterEquipments> getUuidFilterEquipmentsMap(Network network, Map<UUID, String> filters) {
+        String workingVariantId = network.getVariantManager().getWorkingVariantId();
+        UUID uuid = ((NetworkImpl) network).getUuid();
+        return exportFilters(new ArrayList<>(filters.keySet()), uuid, workingVariantId)
+                .stream()
+                .peek(t -> t.setFilterName(filters.get(t.getFilterId())))
```
EstherDarkish, line 78: Is peek appropriate ? Shouldn't it only be used for debugging ?

YenguiSeddik: As mentioned here: https://www.baeldung.com/java-streams-peek-api, i think the use of peek is discouraged when there is no terminal operation. In this case we have one. I used it here because we want to change the inner state of element without changing the returned element itself.

EstherDarkish (approved):

src/main/java/org/gridsuite/modification/server/modifications/ByFormulaModification.java
YenguiSeddik: fixed

src/main/java/org/gridsuite/modification/server/modifications/ByFormulaModification.java
YenguiSeddik: fixed

anistouri (approved):

PR #366, opened 2023-11-14
Aggregate functional logs

Tristan-WorkGH:
Add resume logs for functional logs in the process of voltage init calc.

PR #373, opened 2023-11-16
Fix reporter "id" attribute name

Tristan-WorkGH:
Fix for #366:

Fix attribute `id` use in name of `Reporter`, which is already used by report-server for gridstudy-app (ie. [ReportService.java#L145](https://github.com/gridsuite/report-server/blob/f5d79efe9ddea8e72d1ac64e9430036341f1395a/src/main/java/org/gridsuite/report/server/ReportService.java#L145)).

image


achour94 (approved):

EstherDarkish (approved):
Code OK, tests OK

PR #378, opened 2023-11-22
fix return impacts when active power control modification

achour94:


src/main/java/org/gridsuite/modification/server/modifications/ModificationUtils.java
```diff
@@ -765,37 +765,47 @@ public Reporter modifyActivePowerControlAttributes(ActivePowerControl<?> activeP
...
+            if (droopInfo != null) {
+                activePowerControlAdder
+                        .withDroop(droop);
+                reports.add(ModificationUtils.getInstance().buildModificationReport(Double.NaN,
+                        droop,
+                        "Droop"));
+            } else {
+                activePowerControlAdder
+                        .withDroop(Double.NaN);
+            }
```
AbdelHedhili, line 805: isn't this slightly better ? 
```
activePowerControlAdder.withDroop(droop);
if (droopInfo != null) {            
    reports.add(ModificationUtils.getInstance().buildModificationReport(Double.NaN, droop, "Droop"));
}
```

src/test/java/org/gridsuite/modification/server/modifications/BatteryModificationTest.java
```diff
@@ -217,6 +218,40 @@ public void testDroopUnchanged() throws Exception {
...
+        modificationToCreateJson = mapper.writeValueAsString(batteryModificationInfos);
+        mvcResult = mockMvc.perform(post(getNetworkModificationUri()).content(modificationToCreateJson).contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk()).andReturn();
+        //check impacts
+        resultAsString = mvcResult.getResponse().getContentAsString();
+        networkModificationResult = mapper.readValue(resultAsString, NetworkModificationResult.class);
+        assertEquals(1, networkModificationResult.getNetworkImpacts().size());
+        assertEquals(1, networkModificationResult.getImpactedSubstationsIds().size());
+        assertEquals("[s2]", networkModificationResult.getImpactedSubstationsIds().toString());
+
```
Meklo, line 252: Wouldn't it be interesting to also check the battery ActivePowerControl values after the modifications ?

achour94: this is not the purpose of this PR, the test of modification is done in the other methods

AbdelHedhili (approved):

src/test/java/org/gridsuite/modification/server/modifications/BatteryModificationTest.java
Meklo: Fair enough

PR #385, opened 2023-11-28
change order of check on pmin and pmax

ghazwarhili:


src/main/java/org/gridsuite/modification/server/modifications/GeneratorModification.java
```diff
@@ -170,9 +170,17 @@ private void modifyGeneratorReactiveCapabilityCurvePoints(GeneratorModificationI
     private Reporter modifyGeneratorActiveLimitsAttributes(GeneratorModificationInfos modificationInfos,
                                                            Generator generator, Reporter subReporter) {
         Reporter subReporterLimits = null;
+        Report reportMaxActivePower;
+        Report reportMinActivePower;
 
-        Report reportMaxActivePower = ModificationUtils.getInstance().applyElementaryModificationsAndReturnReport(generator::setMaxP, generator::getMaxP, modificationInfos.getMaxActivePower(), "Max active power");
-        Report reportMinActivePower = ModificationUtils.getInstance().applyElementaryModificationsAndReturnReport(generator::setMinP, generator::getMinP, modificationInfos.getMinActivePower(), "Min active power");
+        if (modificationInfos.getMaxActivePower().getValue() > generator.getMinP()) {
```
FranckLecuyer, line 176: modificationInfos.getMaxActivePower() can be null ...

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/modification/server/modifications/GeneratorModification.java
souissimai: should we also check if modificationInfos.getMaxActivePower().getValue()>modificationInfos.getMinActivePower().getValue() ?

FranckLecuyer (approved):

src/main/java/org/gridsuite/modification/server/modifications/GeneratorModification.java
```diff
@@ -184,9 +184,17 @@ private void modifyGeneratorReactiveCapabilityCurvePoints(GeneratorModificationI
...
+        if (modificationInfos.getMaxActivePower() != null && modificationInfos.getMaxActivePower().getValue() > generator.getMinP()) {
+            reportMaxActivePower = ModificationUtils.getInstance().applyElementaryModificationsAndReturnReport(generator::setMaxP, generator::getMaxP, modificationInfos.getMaxActivePower(), "Max active power");
+            reportMinActivePower = ModificationUtils.getInstance().applyElementaryModificationsAndReturnReport(generator::setMinP, generator::getMinP, modificationInfos.getMinActivePower(), "Min active power");
+
+        } else {
+            reportMinActivePower = ModificationUtils.getInstance().applyElementaryModificationsAndReturnReport(generator::setMinP, generator::getMinP, modificationInfos.getMinActivePower(), "Min active power");
+            reportMaxActivePower = ModificationUtils.getInstance().applyElementaryModificationsAndReturnReport(generator::setMaxP, generator::getMaxP, modificationInfos.getMaxActivePower(), "Max active power");
+        }
```
Mathieu-Deharbe, line 197: Looks very verbose but I don't have a better idea. So ok.

On the other hand, just in case : 
* lines could be split (I can't even see them full on my computer without scrolling left and right)
* there is a useless empty line in the middle

Mathieu-Deharbe (approved):

PR #411, opened 2024-01-02
Optimize SQL queries when getting tabular modification of generators

flomillot:
Fix the n+1 problem when building a node with generators tabular modification.
We just have to call the right method already implemented.


src/test/java/org/gridsuite/modification/server/modifications/tabularmodifications/TabularGeneratorModificationsTest.java
```diff
@@ -120,6 +120,13 @@ public void testCheckSqlRequestsCount() throws Exception {
                 .andReturn();
         // We check that the request count is not dependent on the number of sub modifications of the tabular modification (the JPA N+1 problem is correctly solved)
         assertSelectCount(3);
+        reset();
+
+        // Now we get back the 2 tabular modifications
```
etiennehomer, line 125: // We get the modification group

etiennehomer (approved):

src/test/java/org/gridsuite/modification/server/modifications/tabularmodifications/TabularGeneratorModificationsTest.java
flomillot: I just precised this to explain why we have 6 SELECT, but OK I change a bit the comment

PR #412, opened 2024-01-03
refactor tabular modification

dbraquart:


src/main/java/org/gridsuite/modification/server/dto/TabularModificationInfos.java
```diff
@@ -35,11 +37,15 @@
...
 @ModificationErrorTypeName("TABULAR_MODIFICATION_ERROR")
 public class TabularModificationInfos extends ModificationInfos {
-    @Schema(description = "Modification type")
+
+    public static final String TABULAR_EQUIPMENT_TYPE = "tabularEquipmentType";
+
+    @Schema(description = "equipment type")
     @NonNull
-    private String modificationType;
+    private IdentifiableType equipmentType;
```
etiennehomer, line 45: I think it's not enough. We can have here types like GENERATOR_MODIFICATION, GENERATOR_CREATION or EQUIPMENT_DELETION

etiennehomer: Using ModificationType enum would be more appropriate

SlimaneAmar: No `ModificationType` represente all modification types of  network
We need only injection modificatio/creation/deletion

SlimaneAmar: operationType { MODIFICATION, CREATION, DELETION } ?

etiennehomer: You want to use 2 fields : IdentifiableType equipmentType +  OperationType operationType ?

dbraquart: Last decision: we use ModificationType enum. No data migration. Same Front-end.

SlimaneAmar (approved):

etiennehomer (approved):
Tests OK

PR #422, opened 2024-01-22
Optim get tabular modif

etiennehomer:


src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -137,61 +141,45 @@ public List<ModificationInfos> getModifications(UUID groupUuid, boolean onlyMeta
...
     @Transactional(readOnly = true)
-    public ModificationInfos getModificationInfo(UUID modificationUuid) {
-        Optional<ModificationEntity> optionalModificationEntity = modificationRepository.findById(modificationUuid);
-        if (!optionalModificationEntity.isPresent()) {
-            throw new NetworkModificationException(MODIFICATION_NOT_FOUND, modificationUuid.toString());
+    public ModificationInfos getModificationInfos(UUID modificationUuid) {
+        return modificationRepository.findById(modificationUuid)
+            .map(this::getModificationInfos)
+            .orElseThrow(() -> new NetworkModificationException(MODIFICATION_NOT_FOUND, modificationUuid.toString()));
+    }
```
thangqp, line 175: IMO, should be as following to avoid throw the same exception when getModificationInfos return a null value and keep the same behavior as the original implementation

public ModificationInfos getModificationInfos(UUID modificationUuid) {
        ModificationEntity modificationEntity = modificationRepository.findById(modificationUuid)
            .orElseThrow(() -> new NetworkModificationException(MODIFICATION_NOT_FOUND, modificationUuid.toString()));

        return  getModificationInfos(modificationEntity);
}

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -137,61 +141,45 @@ public List<ModificationInfos> getModifications(UUID groupUuid, boolean onlyMeta
...
                 .collect(Collectors.toList());
         }
     }
 
     @Transactional(readOnly = true)
-    public ModificationInfos getModificationInfo(UUID modificationUuid) {
-        Optional<ModificationEntity> optionalModificationEntity = modificationRepository.findById(modificationUuid);
-        if (!optionalModificationEntity.isPresent()) {
-            throw new NetworkModificationException(MODIFICATION_NOT_FOUND, modificationUuid.toString());
+    public ModificationInfos getModificationInfos(UUID modificationUuid) {
```
thangqp, line 171: Should be as following to avoid throw the same exception when getModificationInfos return null

`    public ModificationInfos getModificationInfos(UUID modificationUuid) {
        ModificationEntity modificationEntity = modificationRepository.findById(modificationUuid)
            .orElseThrow(() -> new NetworkModificationException(MODIFICATION_NOT_FOUND, modificationUuid.toString()));

        return  getModificationInfos(modificationEntity);
    }`

src/main/java/org/gridsuite/modification/server/repositories/equipmentmodification/BatteryModificationRepository.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.UUID;
+
+/**
+ * @author Etienne Homer <etienne.homer at rte-france.com>
+ */
+@Repository
+public interface BatteryModificationRepository extends JpaRepository<BatteryModificationEntity, UUID>, EagerNetworkModificationRepository<BatteryModificationEntity> {
+
+    @Override
+    @EntityGraph(attributePaths = {"reactiveCapabilityCurvePoints", "properties"}, type = EntityGraph.EntityGraphType.LOAD)
```
klesaulnier, line 24: I thought it was not possible to fetch two distinct collections from the same entity in JPA, see below exception
"org.hibernate.loader.MultipleBagFetchException: cannot simultaneously fetch multiple bags"

src/test/java/org/gridsuite/modification/server/modifications/tabularmodifications/TabularGeneratorModificationsTest.java
```diff
@@ -87,48 +88,50 @@ protected void assertAfterNetworkModificationDeletion() {
...
+    public void testSqlRequestsCountOnGetGroupModifications() throws Exception {
+        createTabularGeneratorModification(1);
+        createTabularGeneratorModification(3);
 
-        // We get the modifications of the group (so the 2 tabular modifications)
-        mockMvc.perform(get("/v1/groups/{groupUuid}/network-modifications", getGroupId()))
-                .andExpect(status().isOk());
-        // We check that the request count is not dependent on the number of sub modifications of the tabular modification (the JPA N+1 problem is correctly solved)
+        reset();
+        ApiUtils.getGroupModifications(mockMvc, getGroupId()); // Getting two tabular modifications with respectively one and three sub-modifications
```
klesaulnier, line 109: this call grows in SQL count if we create more tabular modifications
it might not be linked to this PR though, we might have this behaviour with other modifications
Same for other concerned tests

src/test/java/org/gridsuite/modification/server/modifications/tabularmodifications/TabularBatteryModificationsTest.java
```diff
@@ -97,4 +103,49 @@ protected void testUpdateModificationMessage(ModificationInfos modificationInfos
...
+            .modifications(createBatteryModificationList(qty))
+            .build();
+        return saveModification(tabularModification);
+    }
+
+    private List<ModificationInfos> createBatteryModificationList(int qty) {
+        List<ModificationInfos> modifications = new ArrayList<>();
+        for (int i = 0; i <= qty; i++) {
+            modifications.add(
+                BatteryModificationInfos.builder()
```
klesaulnier, line 142: It could be nice to add some "reactiveCapabilityCurvePoints" to this builder to cover the @EntityGraph usage 
Same for other concerned tests

klesaulnier (commented):
It could be nice to find a way to combine those tests classes
It could make them easier to maintain

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -32,11 +27,8 @@ public interface ModificationRepository extends JpaRepository<ModificationEntity
     @Query(value = "SELECT new ModificationEntity(m.id, m.type, m.date, m.stashed, m.messageType, m.messageValues) FROM ModificationEntity m WHERE m.group.id = ?1 order by m.modificationsOrder")
     List<ModificationEntity> findAllBaseByGroupId(UUID uuid);
 
-    @EntityGraph(attributePaths = {"modifications", "modifications.reactiveCapabilityCurvePoints"}, type = EntityGraph.EntityGraphType.LOAD)
-    Optional<TabularModificationEntity> findAllWithReactiveCapabilityCurvePointsById(UUID id);
-
-    @EntityGraph(attributePaths = {"reactiveCapabilityCurvePoints"}, type = EntityGraph.EntityGraphType.LOAD)
-    Set<GeneratorModificationEntity> findAllReactiveCapabilityCurvePointsByIdIn(List<UUID> ids);
+    @Query(value = "SELECT cast(modifications_id as varchar) from tabular_modification_modifications where tabular_modification_entity_id= ?1", nativeQuery = true)
```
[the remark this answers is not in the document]
TheMaskedTurtle: Fixed, I also put SQL keywords in capital letters to harmonize

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -32,11 +27,8 @@ public interface ModificationRepository extends JpaRepository<ModificationEntity
...
+    List<UUID> findSubModificationsIds(UUID uuid);
```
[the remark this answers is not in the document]
TheMaskedTurtle: As the relation table is not defined explicitly we cannot query it directly through JPA, so either we do it with native query directly on the relation table or we go though JPA but we need to add join because to query on TabularModificationEnity. So, I'd rather keep this solution, though the join solution would be quite equally performant I think.

PR #431, opened 2024-02-07
fix(tabular-modifications): improve performance of copies

TheMaskedTurtle:


src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -27,6 +27,9 @@ public interface ModificationRepository extends JpaRepository<ModificationEntity
     @Query(value = "SELECT new ModificationEntity(m.id, m.type, m.date, m.stashed, m.messageType, m.messageValues) FROM ModificationEntity m WHERE m.group.id = ?1 order by m.modificationsOrder")
     List<ModificationEntity> findAllBaseByGroupId(UUID uuid);
 
+    @Query(value = "SELECT m FROM ModificationEntity m WHERE m.id IN :uuids order by m.modificationsOrder")
+    List<ModificationEntity> findAllByIdIn(List<UUID> uuids);
```
sBouzols, line 31: why not using directly the form `findAllByIdInOrderByModificationsOrder` instead of this JPQL query ?

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
```diff
@@ -281,32 +274,42 @@ private Optional<NetworkModificationResult> saveAndApplyModifications(UUID targe
...
-        List<ModificationEntity> duplicatedModificationsEntities = networkModificationRepository.copyModificationsEntities(originGroupUuid);
-        return saveAndApplyModifications(targetGroupUuid, networkUuid, variantId, reportInfos, duplicatedModificationsEntities);
+    public Optional<NetworkModificationResult> duplicateModifications(UUID targetGroupUuid,
+                                                                      UUID networkUuid, String variantId,
+                                                                      ReportInfos reportInfos,
+                                                                      UUID originGroupUuid) {
+        List<ModificationInfos> modificationInfos = networkModificationRepository.getModificationsInfos(originGroupUuid, false);
+        saveModifications(targetGroupUuid, modificationInfos);
+        return applyModifications(networkUuid, variantId, reportInfos, modificationInfos);
     }
```
thangqp, line 299: @Transactional is removed, do we need rollback if a runtime exception occurs while invoking applyModifications()???

PR #434, opened 2024-02-12
add log warn when connection status cannot be changed

dbraquart:


FranckLecuyer (changes requested):
Can you add test cases where the terminal connection/disconnection fails because of no switch(es) found to do the operation, to improve code coverage ...

FranckLecuyer (approved):

PR #442, opened 2024-02-19
performance: improve tabular modification deletions

TheMaskedTurtle:


src/main/java/org/gridsuite/modification/server/repositories/GeneratorModificationRepository.java
```diff
@@ -26,4 +28,16 @@ public interface GeneratorModificationRepository extends JpaRepository<Generator
...
+    @Modifying
+    @Query(value = "BEGIN;" +
+        "DELETE FROM generator_modification_entity_reactive_capability_curve_points cp WHERE cp.generator_modification_entity_id IN ?1 ;" +
+        "DELETE FROM free_property fp WHERE fp.equipment_modification_id IN ?1 ;" +
+        "DELETE FROM generator_modification WHERE id IN ?1 ;" +
+        "DELETE FROM tabular_modification_modifications WHERE tabular_modification_entity_id = ?2 ;" +
+        "DELETE FROM modification WHERE id IN ?1 ;" +
+        "DELETE FROM tabular_modification WHERE id = ?2 ;" +
+        "DELETE FROM modification WHERE id = ?2 ;" +
+        "COMMIT;", nativeQuery = true)
```
etiennehomer, line 41: `"DELETE FROM free_property fp WHERE fp.equipment_modification_id IN ?1 ;" +`
`"DELETE FROM tabular_modification_modifications WHERE tabular_modification_entity_id = ?2 ;" +`
`"DELETE FROM modification WHERE id IN ?1 ;" +`
`"DELETE FROM tabular_modification WHERE id = ?2 ;" +`
`"DELETE FROM modification WHERE id = ?2 ;"`

will be duplicated for each equipment modification. Could we put it common but keep the transaction ?

src/test/java/org/gridsuite/modification/server/modifications/tabularmodifications/TabularGeneratorModificationsTest.java
```diff
@@ -324,6 +324,237 @@ public void testSqlRequestsCountOnPostNetworkModificationsDuplicate2() throws Ex
         assertTabularModificationsEquals(modifications, idsMapping);
     }
 
+    /*
```
etiennehomer, line 327: Maybe we should move those tests to ModificationRepositoryTest or a new file GeneratorModificationRepositoryTest and make them work without any API calls. Can be done in another PR

src/main/java/org/gridsuite/modification/server/repositories/GeneratorModificationRepository.java
TheMaskedTurtle: We'll see that when we do the generalization ticket, no ?

src/test/java/org/gridsuite/modification/server/modifications/tabularmodifications/TabularGeneratorModificationsTest.java
TheMaskedTurtle: I think it's nice to test that through API calls no ? It's more like integration tests, but at least we're sure we test all DB operations end-to-end while calling the endpoint.

etiennehomer (approved):

etiennehomer (approved):

PR #444, opened 2024-02-20
improve tabular modifications move

dbraquart:


src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -84,53 +85,57 @@ private void saveModificationsNonTransactional(UUID groupUuid, List<? extends Mo
...
-                newDestinationModificationList = updateModificationList(modificationsToMoveUUID, originModifications, destinationModifications, referenceModificationUuid);
-
-                originModificationGroupEntity.setModifications(new ArrayList<>(originModifications.values()));
-                destinationModificationGroupEntity.setModifications(newDestinationModificationList);
+        if (originGroupUuid.equals(destinationGroupUuid)) { // single group case
+            if (referenceModificationUuid != null && !originModifications.containsKey(referenceModificationUuid)) {
+                throw new NetworkModificationException(MOVE_MODIFICATION_ERROR);
+            }
+            // re-organize the list
+            updateModificationsOrder(updateModificationList(modificationsToMoveUUID, originModifications, referenceModificationUuid));
```
etiennehomer, line 100: Here it's a bit weird that updateModificationList() modify the originModifications map AND at the same time, updateModificationList() is mostly used for its returned value.

I feel like this method could return the updated list without modiying the parameters.
And updateModificationList() and insertModifications() could work with lists as parameters and not with a map.

src/test/java/org/gridsuite/modification/server/service/ModificationRepositoryTest.java
```diff
@@ -456,16 +457,20 @@ public void insertModificationTest() {
         var groovyScriptEntity4 = GroovyScriptInfos.builder().script("script4").build().toEntity();
         var groovyScriptEntity5 = GroovyScriptInfos.builder().script("script5").build().toEntity();
         var groovyScriptEntity6 = GroovyScriptInfos.builder().script("scriptSaucisse").build().toEntity();
+        List<ModificationEntity> modificationEntities = List.of(groovyScriptEntity1, groovyScriptEntity2, groovyScriptEntity3, groovyScriptEntity4, groovyScriptEntity5, groovyScriptEntity6);
```
etiennehomer, line 460: That would nice to add a tabular modification with few elements to this list.
And then the test would show that the number of requests is not impacted by a tabular modification and its sub elelements.

etiennehomer (changes requested):

src/test/java/org/gridsuite/modification/server/service/ModificationRepositoryTest.java
dbraquart: yes, done.
And it did not change the counters

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
dbraquart: yes, I tried to change/rename things a bit for clarity.

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
```diff
@@ -218,26 +219,26 @@ public void deleteNetworkModifications(UUID groupUuid, List<UUID> modificationsU
...
-                                                                 UUID before, UUID networkUuid, String variantId,
+                                                                 UUID beforeModificationUuid, UUID networkUuid, String variantId,
                                                                  ReportInfos reportInfos, List<UUID> modificationsToMove,
                                                                  boolean canBuildNode) {
-        List<ModificationInfos> movedModifications = networkModificationRepository.moveModifications(groupUuid, originGroupUuid, modificationsToMove, before)
-            .stream()
-            .filter(m -> !m.getStashed())
-            .map(networkModificationRepository::getModificationInfos)
-            .collect(Collectors.toList());
+        List<ModificationEntity> movedMetadataEntities = networkModificationRepository.moveModifications(groupUuid, originGroupUuid, modificationsToMove, beforeModificationUuid);
```
TheMaskedTurtle, line 225: All the modifications corresponding to the groupUuid should be returned here to perform the application correctly. I see 2 options: 
- We return everything but then I would rater return the ModificationInfos directly
- We don't return anything and the application part of this method handles directly the application, so it retrieves ModificationInfos from groupUuid and then apply them if necessary 

What do you think ?

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -48,4 +50,12 @@ public interface ModificationRepository extends JpaRepository<ModificationEntity
...
 
     Integer countByGroupIdAndStashed(UUID groupId, boolean stashed);
+
+    @Modifying
+    @Query("update ModificationEntity m set m.modificationsOrder = :order where m.id = :id")
+    int setOrderById(@Param("order") int order, @Param("id") UUID id);
+
+    @Modifying
+    @Query("update ModificationEntity m set m.group.id = :newGroup where m.id = :id")
+    int setGroupById(@Param("newGroup") UUID newGroup, @Param("id") UUID id);
```
TheMaskedTurtle, line 60: Did you try to modify modification list at group level and the referenced group at modification level, in entity objects directly ? So that JPA handles the DB changes directly

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -84,54 +85,58 @@ private void saveModificationsNonTransactional(UUID groupUuid, List<? extends Mo
...
+        }
 
-                originModificationGroupEntity.setModifications(new ArrayList<>(originModifications.values()));
-                destinationModificationGroupEntity.setModifications(newDestinationModificationList);
+        if (originGroupUuid.equals(destinationGroupUuid)) { // single group case
+            if (referenceModificationUuid != null && !originModifications.containsKey(referenceModificationUuid)) {
+                throw new NetworkModificationException(MOVE_MODIFICATION_ERROR);
+            }
+            // re-organize the origin list
+            List<ModificationEntity> removedModifications = modificationsToMoveUUID.stream().map(originModifications::remove).toList();
```
TheMaskedTurtle, line 104: rename modificationsToMove ? I think it would make it clearer

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -84,54 +85,58 @@ private void saveModificationsNonTransactional(UUID groupUuid, List<? extends Mo
...
+
+            // load destination entities (metadata light mode)
+            List<ModificationEntity> destinationModificationEntities = getModificationsMetadataEntities(destinationGroupUuid, false);
+            Map<UUID, ModificationEntity> destinationModifications = destinationModificationEntities.stream()
+                    .collect(Collectors.toMap(ModificationEntity::getId, Function.identity(), (x, y) -> y, LinkedHashMap::new));
+            if (referenceModificationUuid != null && !destinationModifications.containsKey(referenceModificationUuid)) {
+                throw new NetworkModificationException(MOVE_MODIFICATION_ERROR);
             }
+            // update origin list
+            List<ModificationEntity> removedModifications = modificationsToMoveUUID.stream().map(originModifications::remove).toList();
```
TheMaskedTurtle, line 119: Same

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -84,54 +85,58 @@ private void saveModificationsNonTransactional(UUID groupUuid, List<? extends Mo
...
-        /* when referenceModification == null we move at the end of list */
-        int index = referenceModificationUuid == null ? newDestinationModificationList.size() : newDestinationModificationList.indexOf(destinationModifications.get(referenceModificationUuid));
-        newDestinationModificationList.addAll(index, movedModifications);
+    private void updateModificationsOrder(Collection<ModificationEntity> modifications) {
+        AtomicInteger index = new AtomicInteger(0);
+        modifications.forEach(m -> this.modificationRepository.setOrderById(index.getAndIncrement(), m.getId()));
+    }
 
-        return newDestinationModificationList;
+    private List<ModificationEntity> createNewModificationList(List<ModificationEntity> modificationsToAdd, Map<UUID, ModificationEntity> modifications, UUID referenceModificationUuid) {
```
TheMaskedTurtle, line 134: rename addModifications or mergeModifications or insertModifications ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -84,54 +85,58 @@ private void saveModificationsNonTransactional(UUID groupUuid, List<? extends Mo
...
     }
 
-    private List<ModificationEntity> updateModificationList(List<UUID> modificationsToMoveUuid, Map<UUID, ModificationEntity> originModifications, Map<UUID, ModificationEntity> destinationModifications, UUID referenceModificationUuid) {
-        List<ModificationEntity> movedModifications = modificationsToMoveUuid.stream().map(originModifications::remove).collect(Collectors.toList());
-
-        List<ModificationEntity> newDestinationModificationList = new ArrayList<>(destinationModifications.values());
-        /* when referenceModification == null we move at the end of list */
-        int index = referenceModificationUuid == null ? newDestinationModificationList.size() : newDestinationModificationList.indexOf(destinationModifications.get(referenceModificationUuid));
-        newDestinationModificationList.addAll(index, movedModifications);
+    private void updateModificationsOrder(Collection<ModificationEntity> modifications) {
```
etiennehomer, line 129: You should use a List here. A collection can be unordered

etiennehomer (changes requested):

src/test/java/org/gridsuite/modification/server/service/ModificationRepositoryTest.java
```diff
@@ -449,23 +451,40 @@ public void testLineCreation() {
     }
 
     @Test
-    public void insertModificationTest() {
+    public void testMoveModificationInSameGroup() {
```
TheMaskedTurtle, line 454: That's quite a lot of DB statements :/ There is no way to by-pass this temporary table creation ? Maybe using JPA properties directly (previous comment) will do the trick, it could be nice to test this solution.

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
```diff
@@ -664,8 +630,68 @@ public void createGeneratorWithStartup() throws Exception {
     }
 
     @Test
-    public void testMoveModificationWithoutOrigin() throws Exception {
-        testMoveModification(null, null);
+    public void testMoveModificationInSameGroup() throws Exception {
```
TheMaskedTurtle, line 633: Is there a way to test that modifications are applied correctly on the network afterward ? Because I think this part of the code is not tested for now

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -84,54 +85,58 @@ private void saveModificationsNonTransactional(UUID groupUuid, List<? extends Mo
...
+                throw new NetworkModificationException(MOVE_MODIFICATION_ERROR);
             }
+            // update origin list
+            List<ModificationEntity> removedModifications = modificationsToMoveUUID.stream().map(originModifications::remove).toList();
+            updateModificationsOrder(originModifications.values());
+            // update destination list
+            updateModificationsOrder(createNewModificationList(removedModifications, destinationModifications, referenceModificationUuid));
+            // change the owner group for moved modifications
+            modificationsToMoveUUID.forEach(modifId -> this.modificationRepository.setGroupById(destinationGroupUuid, modifId));
+            return removedModifications;
```
TheMaskedTurtle, line 125: Same, I think this is not the previous behavior.

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
TheMaskedTurtle: I just saw you renamed it from previous review ^^ I just think `createNewModificationList` is a bit too generic

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -84,54 +85,58 @@ private void saveModificationsNonTransactional(UUID groupUuid, List<? extends Mo
...
-                originModificationGroupEntity.setModifications(new ArrayList<>(originModifications.values()));
-                destinationModificationGroupEntity.setModifications(newDestinationModificationList);
+        if (originGroupUuid.equals(destinationGroupUuid)) { // single group case
+            if (referenceModificationUuid != null && !originModifications.containsKey(referenceModificationUuid)) {
+                throw new NetworkModificationException(MOVE_MODIFICATION_ERROR);
+            }
+            // re-organize the origin list
+            List<ModificationEntity> removedModifications = modificationsToMoveUUID.stream().map(originModifications::remove).toList();
+            updateModificationsOrder(createNewModificationList(removedModifications, originModifications, referenceModificationUuid));
+            return List.of();
```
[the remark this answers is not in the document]
dbraquart: not really beacuse we now canBuild is false in this case. But yes, it's better to return the moved modifications.

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
dbraquart: as said before, this PR implements the right behaviour

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
dbraquart: hum I dont know if we will agree on this. newListWithInsertedModifications() ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
dbraquart: this function is gone actually

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
dbraquart: I revert these changes: no need anymore

src/test/java/org/gridsuite/modification/server/service/ModificationRepositoryTest.java
dbraquart: I dont update order/group manually anymore. JPA manages everything, and there is no more tmp table, then no more insert/remove.

TheMaskedTurtle (approved):
Nice :)

etiennehomer (approved):

PR #447, opened 2024-02-27
 Homogenize danglingline attribute

souissimai:


Mathieu-Deharbe (approved):
A full PR for this seems a bit extreme but ok let's go full homogenization.

AAJELLAL (approved):

PR #453, opened 2024-03-12
make sure order column starts from 0 in group modifications list

dbraquart:


src/main/java/org/gridsuite/modification/server/entities/ModificationGroupEntity.java
```diff
@@ -39,8 +39,14 @@ public ModificationGroupEntity(UUID uuid) {
     }
 
     public void addModification(ModificationEntity modification) {
-        modifications.add(modification);
-        modification.setGroup(this);
+        if (modifications.isEmpty()) {
+            List<ModificationEntity> newList = new ArrayList<>();
+            newList.add(modification);
+            setModifications(newList);
```
SlimaneAmar, line 45: `setModifications(List.of(modification)) `?

dbraquart: This was my first shot. But I we pass an unmutable collection like List(), we got an exception when we try to add a second element to the modifications list.

src/main/java/org/gridsuite/modification/server/entities/ModificationGroupEntity.java
```diff
@@ -39,8 +39,14 @@ public ModificationGroupEntity(UUID uuid) {
     }
 
     public void addModification(ModificationEntity modification) {
-        modifications.add(modification);
-        modification.setGroup(this);
+        if (modifications.isEmpty()) {
```
TheMaskedTurtle, line 42: Maybe add a comment to explain we have to do that to maintain proper order with @OrderColumn annotation

TheMaskedTurtle (approved):

src/main/java/org/gridsuite/modification/server/entities/ModificationGroupEntity.java
SlimaneAmar: Arrays.asList

dbraquart: Arrays.asList also returns "a fixed-size list"

dbraquart: same issue with:
 List<ModificationEntity> newList = Arrays.asList(modification);

dbraquart: sonar doesnt like it ;-)
_Double Brace Initialization should not be used_

src/main/java/org/gridsuite/modification/server/entities/ModificationGroupEntity.java
dbraquart: done

etiennehomer (approved):

PR #463, opened 2024-04-05
Identification and first fixes of log overwriting issues.

EstherDarkish:
In a second PR, the identified but not fixed issues will be adressed.

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/modifications/BatteryCreation.java
+++ b/src/main/java/org/gridsuite/modification/server/modifications/BatteryCreation.java
@@ -219,8 +219,9 @@ public class BatteryCreation extends AbstractModification {
             } catch (PowsyblException e) {
                 activePowerRegulationReports.add(Report.builder()
                         .withKey("ActivePowerExtensionAddError")
-                        .withDefaultMessage("cannot add active power extension on battery with id=${id} : " + e.getMessage())
+                        .withDefaultMessage("cannot add active power extension on battery with id=${id} : ${message}")
                         .withValue("id", batteryCreationInfos.getEquipmentId())
+                        .withValue("message", e.getMessage())
                         .withSeverity(TypedValue.ERROR_SEVERITY)
                         .build());
             }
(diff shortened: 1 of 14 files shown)


src/main/java/org/gridsuite/modification/server/modifications/BatteryCreation.java
```diff
@@ -219,8 +219,9 @@ private void createBatteryActivePowerControl(BatteryCreationInfos batteryCreatio
             } catch (PowsyblException e) {
                 activePowerRegulationReports.add(Report.builder()
                         .withKey("ActivePowerExtensionAddError")
-                        .withDefaultMessage("cannot add active power extension on battery with id=${id} : " + e.getMessage())
+                        .withDefaultMessage("cannot add active power extension on battery with id=${id} : ${message}")
                         .withValue("id", batteryCreationInfos.getEquipmentId())
+                        .withValue("message", e.getMessage())
```
dbraquart, line 224: I think as soon as you have a parameter ({id} here) in the message, it is ok

src/main/java/org/gridsuite/modification/server/modifications/OperatingStatusModification.java
```diff
@@ -151,9 +154,10 @@ private void applyEnergiseEquipmentEnd(Reporter subReporter, Identifiable<?> equ
 
         subReporter.report(Report.builder()
                 .withKey("energise" + equipmentType + "EndApplied")
-                .withDefaultMessage(equipmentType + " ${id} (id) : energise the side ${side} applied")
+                .withDefaultMessage("${equipmentType} ${id} (id) : energise the side ${side} applied")
                 .withValue("id", equipment.getId())
                 .withValue("side", side.name())
+                .withValue("equipmentType", equipmentType)
```
dbraquart, line 160: would be better if the withValue statements follow the same left->right order

dbraquart (approved):
code: ok (minor remarks)
tests: ok (partial)

src/main/java/org/gridsuite/modification/server/modifications/BatteryCreation.java
EstherDarkish: Every variable needs to be in the .withValue clause

src/main/java/org/gridsuite/modification/server/modifications/OperatingStatusModification.java
EstherDarkish: Fixed

ghazwarhili (approved):
code => ok

PR #468, opened 2024-04-19
fix: add deleted equipment in tombstoned

AbdelHedhili:


src/test/java/org/gridsuite/modification/server/service/ModificationElasticsearchTest.java
```diff
@@ -103,6 +106,30 @@ public void testModificationsToImpactElasticsearch() throws Exception {
         assertEquals("v1load_newname", equipmentInfosRepository.findByIdInAndNetworkUuidAndVariantId(List.of("v1Load"), NETWORK_UUID, NEW_VARIANT).get(0).getName());
     }
 
+    @Test
+    public void testModifyGenThenDeleteVl() throws Exception {
```
SlimaneAmar, line 110: Move this test in a file with inteagrtion tests

AbdelHedhili: As discussed it is already how this class is used, so leaving it as is for now

SlimaneAmar (approved):

PR #483, opened 2024-05-16
Exclude switch equipments from elasticsearch indexation

achour94:


src/main/java/org/gridsuite/modification/server/elasticsearch/EquipmentInfosService.java
```diff
@@ -33,13 +35,19 @@ public class EquipmentInfosService {
...
     public void addAllEquipmentInfos(@NonNull final List<EquipmentInfos> equipmentsInfos) {
-        Lists.partition(equipmentsInfos, partitionSize)
+        // get only equipments allowed to be indexed
+        List<EquipmentInfos> filteredEquipmentsInfos = equipmentsInfos.stream()
+                .filter(equipmentInfos -> !EXCLUDED_TYPES_FOR_INDEXING.contains(equipmentInfos.getType()))
+                .toList();
+        Lists.partition(filteredEquipmentsInfos, partitionSize)
                 .parallelStream()
                 .forEach(equipmentInfosRepository::saveAll);
     }
```
FranckLecuyer, line 53: Add this filter to exclude switches also in addAllTombstonedEquipmentInfos ?
because addAllTombstonedEquipmentInfos is also called in cloneVariantModifications

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/modification/server/elasticsearch/EquipmentInfosService.java
achour94: In cloneVariantModifications, we clone the equipment that is already indexed in 'tombstonedEquipment'. However, since we filter the equipment types during the initial indexing, we will not have them here

FranckLecuyer: Yes, but what for existing network variants where switches have already been indexed before in tombstonedEquipment ?

achour94: It will be more easy to just clean elasticsearch indexes, so we will do that :)

FranckLecuyer: yes

FranckLecuyer (approved):

anistouri (approved):

PR #496, opened 2024-07-08
Use voltageLevelId to improve getBus() lookup time

antoinebhs:
While doing tests with the profiler on VoltageInitModification, I found an issue on the getBus(id) for the VoltageInit modification. As buses are not cache, it takes forever to retrieve the buses using the lookup getBus(id) at each iteration.
Before this PR the time taken to apply a voltage-init modification was about 30s, after it will be around 5-10s.


jonenst:
good diagnosis, but maybe we need a more general solution ?

SlimaneAmar (approved):

FranckLecuyer (approved):

PR #507, opened 2024-07-30
Fix generation dispatch log

ayolab:


dbraquart (approved):
code: ok
tests: ok
Ex: _WARN Cannot find the following generators toto, titi in filter bad_gene_

PR #509, opened 2024-08-01
Modify generation dispatch logs related to not found generators

ayolab:
a synthetic message: WARN - Cannot find N generators in filter filterName.
and a TRACE message for each not found generator: TRACE - Cannot find generator notFoundGeneratorId in filter filterName.


dbraquart (approved):
code: ok
tests: ok
Ex:
```
WARN Cannot find 2 generators in filter bad_gene
TRACE Cannot find generator toto in filter bad_gene
TRACE Cannot find generator titi in filter bad_gene
```

PR #523, opened 2024-08-27
Check null values for droop and P0 (2e patch)

thangqp:
A PR completed for the previous PR: https://github.com/gridsuite/network-modification-server/pull/512

When extension is not yet created:

- If all fields are provided => OK
- If one of fields is provided and one of others fields not provided => NOK
- If none of fields is provided => OK

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/modifications/VscCreation.java
+++ b/src/main/java/org/gridsuite/modification/server/modifications/VscCreation.java
@@ -23,7 +23,6 @@ import java.util.List;
 import java.util.Map;
 
 import static org.gridsuite.modification.server.NetworkModificationException.Type.*;
-import static org.gridsuite.modification.server.modifications.VscModification.DROOP_ACTIVE_POWER_CONTROL_P0_DROOP_REQUIRED_ERROR_MSG;
 
 /**
  * @author Seddik Yengui <seddik.yengui at rte-france.com>
@@ -33,6 +32,7 @@ public class VscCreation extends AbstractModification {
 
     public static final String VSC_SETPOINTS = "vscSetPoints";
     public static final String VSC_CHARACTERISTICS = "vscCharacteristics";
+    public static final String DROOP_P0_REQUIRED_ERROR_MSG = "Droop and P0 are both required when angle droop active power control is activated";
 
     private final VscCreationInfos modificationInfos;
 
@@ -59,8 +59,7 @@ public class VscCreation extends AbstractModification {
 
         // enable the extension => should verify whether all fields have been filled
         if (modificationInfos.getDroop() == null || modificationInfos.getP0() == null) {
-            throw new NetworkModificationException(WRONG_HVDC_ANGLE_DROOP_ACTIVE_POWER_CONTROL,
-                    String.format(DROOP_ACTIVE_POWER_CONTROL_P0_DROOP_REQUIRED_ERROR_MSG));
+            throw new NetworkModificationException(WRONG_HVDC_ANGLE_DROOP_ACTIVE_POWER_CONTROL, DROOP_P0_REQUIRED_ERROR_MSG);
         }
     }
 
--- a/src/main/java/org/gridsuite/modification/server/modifications/VscCreation.java
+++ b/src/main/java/org/gridsuite/modification/server/modifications/VscCreation.java
@@ -24,8 +24,8 @@ import java.util.Map;
 
 import static org.gridsuite.modification.server.NetworkModificationException.Type.CREATE_VSC_ERROR;
 import static org.gridsuite.modification.server.NetworkModificationException.Type.HVDC_LINE_ALREADY_EXISTS;
-import static org.gridsuite.modification.server.modifications.ModificationUtils.checkHvdcDroop;
-import static org.gridsuite.modification.server.modifications.ModificationUtils.shouldCreateHvdcDroopActivePowerControlExtension;
+import static org.gridsuite.modification.server.modifications.VscModification.checkHvdcDroop;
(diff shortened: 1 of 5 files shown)


Mathieu-Deharbe (changes requested):
I quote Thibault : 
_the switch off and the k and P0 fields filled means that an extension is created (with AC emulation disabled)_
This is not the case in this PR (in this situation you don't create the extension)

thangqp:
> I quote Thibault : _the switch off and the k and P0 fields filled means that an extension is created (with AC emulation disabled)_ This is not the case in this PR (in this situation you don't create the extension)
DONE
I created a share method in ModificationUtils which is used in both creation and modification, see https://github.com/gridsuite/network-modification-server/pull/523/files#diff-d9597ed9d121f52bb8e4e81e3783e97fc6d80bdf40ec4d6bf7f77686f1b1bf9aR1310

src/main/java/org/gridsuite/modification/server/modifications/VscModification.java
```diff
@@ -34,15 +35,30 @@ public class VscModification extends AbstractModification {
...
+    public static void checkDroop(boolean isPresentAngleDroopActivePowerControl, boolean isPresentDroop, boolean isPresentP0) {
+        // all fields should be provided => OK extension will be created
+        if (isPresentAngleDroopActivePowerControl && isPresentDroop && isPresentP0) {
+            return;
+        }
+        // at least one field is provided but not for others => NOT OK
+        // all fields are not provided => OK extension will not be created
+        if (isPresentAngleDroopActivePowerControl || isPresentDroop || isPresentP0) {
+            throw new NetworkModificationException(WRONG_HVDC_ANGLE_DROOP_ACTIVE_POWER_CONTROL, ACTIVE_POWER_CONTROL_DROOP_P0_REQUIRED_ERROR_MSG);
+        }
```
Mathieu-Deharbe, line 55: This test doesn't work in the case of the Vsc creation : if you don't check the "Emulation AC" in a VSC creation interface it will still be present. Just `false`. So `isPresentAngleDroopActivePowerControl` is true even if `AngleDroopActivePowerControl` is false and an exception is thrown.

In other words, this check makes it impossible to create a VSC without ActivePowerControl extension (I tried).

I think putting the creation and modification checks in the same function is a good try but dangerous because there are often many small differences like this one.

Mathieu-Deharbe (changes requested):

src/main/java/org/gridsuite/modification/server/modifications/VscModification.java
thangqp: DONE!

I added a particular case which accepts Droop and P0 not provided when not enabling extension for creation. I removed also the static method checkDroop. Each one has its own checkDroop method even they have similar implementation.

src/test/java/org/gridsuite/modification/server/modifications/VscCreationTest.java
```diff
@@ -312,22 +311,67 @@ public void testCreateWithoutEnablingDroopPowerControl() throws Exception {
...
-        vscCreationInfos.setAngleDroopActivePowerControl(true);
+        vscCreationInfos.setAngleDroopActivePowerControl(false);
         vscCreationInfos.setDroop(null);
         vscCreationInfos.setP0(null);
+        String vscCreationInfosJson = mapper.writeValueAsString(vscCreationInfos);
+        mockMvc.perform(post(getNetworkModificationUri()).content(vscCreationInfosJson).contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk());
+        assertThat(getNetwork().getHvdcLine("vsc1")).isNotNull();
+        HvdcLine hvdcLine = getNetwork().getHvdcLine("vsc1");
+        assertThat(hvdcLine).isNotNull();
```
Mathieu-Deharbe, line 332: Aren't you checking the same thing twice ?

Mathieu-Deharbe (approved):

src/main/java/org/gridsuite/modification/server/modifications/VscCreation.java
```diff
@@ -52,16 +51,22 @@ public void check(Network network) throws NetworkModificationException {
...
-
-        // enable the extension => should verify whether all fields have been filled
-        if (modificationInfos.getDroop() == null || modificationInfos.getP0() == null) {
-            throw new NetworkModificationException(WRONG_HVDC_ANGLE_DROOP_ACTIVE_POWER_CONTROL,
-                    String.format(DROOP_ACTIVE_POWER_CONTROL_P0_DROOP_REQUIRED_ERROR_MSG));
+        // particular case, not enabling extension and others fields are not provided => OK extension will not be created
+        if (Boolean.FALSE.equals(modificationInfos.getAngleDroopActivePowerControl()) && !isPresentDroop && !isPresentP0) {
+            return;
+        }
+        // at least one field is provided but not for others => NOT OK
```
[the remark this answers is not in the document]
thangqp: DONE! 
Corrected also in modification

src/main/java/org/gridsuite/modification/server/modifications/VscCreation.java
```diff
@@ -52,16 +51,22 @@ public void check(Network network) throws NetworkModificationException {
...
+        if (isPresentAngleDroopActivePowerControl || isPresentDroop || isPresentP0) {
+            throw new NetworkModificationException(WRONG_HVDC_ANGLE_DROOP_ACTIVE_POWER_CONTROL, VscModification.ACTIVE_POWER_CONTROL_DROOP_P0_REQUIRED_ERROR_MSG);
         }
+        // otherwise, i.e. all fields are not provided => OK extension will not be created
```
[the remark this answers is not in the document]
thangqp: DONE! 
Corrected also in modification

PR #529, opened 2024-09-04
Clean logs to avoid duplicates between containers and logs

dbraquart:
With https://github.com/gridsuite/gridstudy-app/pull/2234 we display all report nodes (report tree on left side) also as logs (messages on right side).

Then we have to remove all logs in the back created only to show these report nodes.


FranckLecuyer (approved):
Code review: OK
Tests: OK

PR #536, opened 2024-09-16
fix connectable position adder attributes

ghazwarhili:


EtienneLt (changes requested):
if (connectionNameReport != null || connectionDirectionReport != null || connectionPositionReport != null) {
   adder.add();
}

src/main/java/org/gridsuite/modification/server/modifications/ModificationUtils.java
```diff
@@ -682,10 +684,8 @@ private void createNewConnectivityPosition(ConnectablePositionAdder<?> adder,
                 CONNECTION_POSITION_FIELD_NAME);
         if (connectionPositionReport != null) {
             reports.add(connectionPositionReport);
+            adder.add();
         }
-
-        // Finalize by adding the feeder
-        adder.add();
```
EtienneLt, line 688: `if (connectionNameReport != null || connectionDirectionReport != null || connectionPositionReport != null) {
   adder.add();
}`

src/main/java/org/gridsuite/modification/server/modifications/ModificationUtils.java
```diff
@@ -664,6 +664,7 @@ private void createNewConnectivityPosition(ConnectablePositionAdder<?> adder,
                 CONNECTION_NAME_FIELD_NAME);
         if (connectionNameReport != null) {
             reports.add(connectionNameReport);
+            adder.add();
```
EtienneLt, line 667: it will add several times the extension if several attributes are not null

EtienneLt (changes requested):

EtienneLt (approved):

PR #560, opened 2024-11-06
fix the get susceptance value per section

ghazwarhili:


EtienneLt (approved):

src/main/java/org/gridsuite/modification/server/modifications/ShuntCompensatorModification.java
```diff
@@ -91,7 +91,7 @@ public static void modifyMaximumSectionCount(AttributeModification<Integer> maxi
         if (maximumSectionCountModif != null) {
             var maximumSectionCount = maximumSectionCountModif.getValue();
             if (maxSusceptance == null && maxQAtNominalV == null) {
-                model.setBPerSection(model.getBPerSection() * shuntCompensator.getMaximumSectionCount() / maximumSectionCount);
+                model.setBPerSection((model.getBPerSection() * shuntCompensator.getMaximumSectionCount()) / maximumSectionCount);
```
EtienneLt, line 94: not needed I think

PR #568, opened 2024-12-03
remove previous values from table

ghazwarhili:


src/main/java/org/gridsuite/modification/server/entities/equipment/creation/ReactiveCapabilityCurveCreationEmbeddable.java
```diff
@@ -31,4 +34,18 @@ public class ReactiveCapabilityCurveCreationEmbeddable {
 
     @Column
     private Double p;
+
```
etiennehomer, line 37: class to be renamed to ReactiveCapabilityCurveEmbeddable ?

src/main/java/org/gridsuite/modification/server/entities/equipment/modification/BatteryModificationEntity.java
```diff
@@ -81,7 +81,7 @@ public class BatteryModificationEntity extends InjectionModificationEntity {
 
     @ElementCollection
     @CollectionTable
-    private List<ReactiveCapabilityCurveModificationEmbeddable> reactiveCapabilityCurvePoints;
+    private List<ReactiveCapabilityCurveCreationEmbeddable> reactiveCapabilityCurvePoints;
```
etiennehomer, line 84: You need to generate a liquibase changeset as you have fewer columns now

etiennehomer (changes requested):

PR #580, opened 2024-12-19
[WIP] Fix filter deletion error in ES

antoinebhs:
It seems to lead to issues if you have modifications that affect a lot of identifiables (voltage init + generation dispatch + ...).
If there are more than 10000 modified identifiables, you can't retrieve the results. You'll get an error about the max result window (https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html, see index.max_result_window)
I revert it for now. Later, we may need to batch the find instead... Or find a better solution.

        "reason" : "Result window is too large, from + size must be less than or equal to: [10000] but was [12973]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."


antoinebhs:
Already merged in another PR

PR #585, opened 2024-12-30
Add a metric for network modifications in the thread pool

SlimaneAmar:


pom.xml
```diff
@@ -174,6 +174,10 @@
             <groupId>org.springframework.cloud</groupId>
             <artifactId>spring-cloud-stream</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-core</artifactId>
+        </dependency>
```
antoinebhs, line 180: In some services we have

            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-actuator</artifactId>

with compile scope instead (for example in study-notification-server, loadflow-server), I'm not sure what's best.
We should probably follow that? Same for gateway probably...

src/main/java/org/gridsuite/modification/server/service/NetworkModificationExecutionService.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.function.Supplier;
+
+/**
+ * @author Slimane Amar <slimane.amar at rte-france.com>
+ */
+@Service
+public class NetworkModificationExecutionService {
```
antoinebhs, line 24: Same for the name, the execution service is special for concurrent-applications, I'm not sure we should have a generic name here?

src/main/java/org/gridsuite/modification/server/service/NetworkModificationObserver.java
```diff
@@ -26,4 +39,11 @@ private Observation createObservation(String name, ModificationType modification
             .lowCardinalityKeyValue(MODIFICATION_TYPE_TAG_NAME, modificationType.name());
     }
 
+    public void createThreadPoolMetric(ThreadPoolExecutor threadPoolExecutor) {
+        MultiGauge multiGauge = MultiGauge.builder(TASK_POOL_METER_NAME).description("The number of large network modifications (tasks) in the thread pool").register(meterRegistry);
+        multiGauge.register(List.of(
+            MultiGauge.Row.of(Tags.of(TASK_TYPE_TAG_NAME, TASK_TYPE_TAG_VALUE_CURRENT), threadPoolExecutor::getActiveCount),
+            MultiGauge.Row.of(Tags.of(TASK_TYPE_TAG_NAME, TASK_TYPE_TAG_VALUE_PENDING), () -> threadPoolExecutor.getQueue().size())
+        ));
+    }
```
antoinebhs, line 48: I'm not sure to understand why we need a multigauge here as the number of tags is constant and equal to 2. From my understanding multigauge are useful if the number of tags vary.
Here, it seems simpler to have two gauges (one for current, one for pending)?

antoinebhs (commented):
Some code comments, will test tomorrow

pom.xml
SlimaneAmar: Done

src/main/java/org/gridsuite/modification/server/service/NetworkModificationExecutionService.java
SlimaneAmar: Wiith Jauge.builder we can't have a value supplier different for different tags !

src/main/java/org/gridsuite/modification/server/service/NetworkModificationObserver.java
SlimaneAmar: Wiith Jauge.builder we can't have a value supplier different for different tags !

src/main/java/org/gridsuite/modification/server/service/NetworkModificationExecutionService.java
SlimaneAmar: LargeNetworkModificationExecutionService ?

src/main/java/org/gridsuite/modification/server/service/NetworkModificationObserver.java
antoinebhs:
```suggestion
    public void createThreadPoolMetric(ThreadPoolExecutor threadPoolExecutor) {
        Gauge.builder(TASK_POOL_METER_NAME + TASK_TYPE_TAG_VALUE_CURRENT, threadPoolExecutor, ThreadPoolExecutor::getActiveCount)
                .description("The number of active diagram generation tasks in the thread pool")
                .tag(TASK_TYPE_TAG_NAME, TASK_TYPE_TAG_VALUE_CURRENT)
                .register(meterRegistry);

        Gauge.builder(TASK_POOL_METER_NAME + TASK_TYPE_TAG_VALUE_PENDING, threadPoolExecutor, executor -> executor.getQueue().size())
                .description("The number of pending diagram generation tasks in the thread pool")
                .tag(TASK_TYPE_TAG_NAME, TASK_TYPE_TAG_VALUE_PENDING)
                .register(meterRegistry);
    }
```
this does not work ?

SlimaneAmar: Done

antoinebhs (approved):
Code OK

jonenst (approved):

PR #587, opened 2025-01-13
Rename composite modification methods

SlimaneAmar:


src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
```diff
@@ -233,26 +226,14 @@ public void deleteNetworkModifications(UUID groupUuid, List<UUID> modificationsU
...
-                List<ModificationInfos> movedModifications = movedEntities.stream()
-                        .map(networkModificationRepository::getModificationInfos).toList();
-                return Optional.of(modificationApplicator.applyModifications(
-                        movedModifications,
-                        networkInfos,
-                        reportInfos));
-            }
-        }
-        return Optional.empty();
+        return applyModifications ? applyModifications(networkUuid, variantId, reportInfos, modificationInfos) : Optional.empty();
```
dbraquart, line 236: I remember I made these changes above (lines 240->255) to optimize the move operation specially with Tabular modifications.
Copying the entities are very fast, but converting entities->dto is very slow. So the idea was to delay this conversion at the last time (just before we are sure to apply).

If we want to keep good performances with Tabular modifications (in the case we dont apply), we cannot merge your change above.
I measured: 
- moving a single big Tabular modification (on all lines) at entity level: 250ms
- converting this modification: 4783ms

Imagine if you move several of them (without apply),

dbraquart (changes requested):
code: issue with tabular modification move wihout applying
tests: ok

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
SlimaneAmar: I change the code to delay this conversion at the last time

dbraquart (approved):

PR #597, opened 2025-02-10
add report node in network to get regulating point logs

EtienneLt:


src/main/java/org/gridsuite/modification/server/modifications/NetworkModificationApplicator.java
```diff
@@ -190,6 +190,8 @@ private void apply(AbstractModification modification, Network network, ReportNod
         // init application context
         modification.initApplicationContext(this.filterService);
 
+        network.getReportNodeContext().pushReportNode(subReportNode);
```
antoinebhs, line 193: I would have added this in ByFilterDeletion.java//EquipmentDeletion.java in the library. What's your take on this?
I'm not sure to understand why we need this. Maybe we can have someone else review this part for another point of view?

etiennehomer: Maybe it's the right solution if we want to send back the reports from network-store and powsybl-core. But it's quite a big change and we don't really control the reports that we'll retrieve. 
It's safer to first add it in the modification which we want to retrieve some reports

antoinebhs:
Replaced by https://github.com/gridsuite/network-modification/pull/21

PR #618, opened 2025-04-02
Add measurements in load modification

ayolab:


src/main/java/org/gridsuite/modification/server/entities/equipment/modification/LoadModificationEntity.java
```diff
@@ -50,6 +52,34 @@ public class LoadModificationEntity extends InjectionModificationEntity {
     @Enumerated(EnumType.STRING)
     private OperationType q0Op;
 
+    @Embedded
```
FranckLecuyer, line 55: Put these extra fields in the base class InjectionModificationEntity ? it will then be used for other injection type (generators, shunt compensator ...)
like for lines and two windings transformers : the measurements fields are in BranchModificationEntity ...

FranckLecuyer (changes requested):

FranckLecuyer (approved):

PR #627, opened 2025-04-22
Reindex modifications

klesaulnier:


src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -21,7 +24,10 @@
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
 
-import java.util.*;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
```
antoinebhs, line 30: Revert?

src/main/java/org/gridsuite/modification/server/service/SupervisionService.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Transactional(readOnly = true)
+    public long getModificationsToReindexCount() {
+        return modificationApplicationRepository.count();
+    }
+
+    @Transactional
+    public void reindexAll() {
+        modificationApplicationInfosRepository.deleteAll();
+        modificationApplicationInfosRepository.saveAll(modificationApplicationRepository.findAllWithModificationAndGroup().stream().map(
```
antoinebhs, line 48: Is this OK if there are a lot of modifications to save ? > 10000 ?

https://github.com/gridsuite/network-modification-server/blob/8dd23305731b882a0316c1c648b16acd01e0b8cc/src/main/java/org/gridsuite/modification/server/elasticsearch/EquipmentInfosService.java#L62-L66

antoinebhs: to test: > 10000 modif

antoinebhs:
I tested with 1 000 000 modifications and I get Java Heap Space because we load in memory the million entities and try to save them in one batch. 

Can we batch the reindexation to avoid that ? In other cases we batch by study for example to keep a low number of stuff to load/save. Maybe we can batch by networkUuid here? I think we tried to avoid very long request in admin scripts up to now.

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
klesaulnier: fixed

PR #669, opened 2025-08-05
Use correct builder for limit sets tabular modifications

Meklo:


dbraquart (approved):
code: ok
test: ok (Line + 2WT)

PR #685, opened 2025-09-08
unsettable fields in voltage level filter modification

Mathieu-Deharbe:


src/main/java/org/gridsuite/modification/server/entities/equipment/modification/byfilter/assignment/AssignmentEntity.java
```diff
@@ -61,10 +61,10 @@ public AssignmentInfos<?> toAssignmentInfos() {
                 .value(Boolean.valueOf(value))
                 .build();
             case INTEGER -> IntegerAssignmentInfos.builder()
-                .value(Integer.valueOf(value))
+                .value(value != null ? Integer.valueOf(value) : null)
```
basseche, line 64: you can also use Optional if you want :

`Optional.ofNullable(value).map(Integer::valueOf).orElse(null)`

PR #690, opened 2025-09-22
Modifications applications: split long running transaction before/after the computations

jonenst:
This requires to reorganize the code and the transactions the fully load the data, hence the changes in NetworkModificationRepository with the new intermediate methods.

Add testInsertCompositeModifications for symmetry with other code paths duplicate/move/ in the switch/case of ModificationController ?action=...

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
+++ b/src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
@@ -106,7 +106,6 @@ public class NetworkModificationRepository {
     }
 
     @Transactional // To have all create in the same transaction (atomic)
-    // TODO Remove transaction when errors will no longer be sent to the front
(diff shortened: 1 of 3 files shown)


src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -117,6 +117,11 @@ public List<ModificationEntity> saveModifications(UUID groupUuid, List<Modificat
     @Transactional // To have all create in the same transaction (atomic)
     // TODO Remove transaction when errors will no longer be sent to the front
     public List<ModificationEntity> saveModificationInfos(UUID groupUuid, List<ModificationInfos> modifications) {
+        return saveModificationInfosNonTransactional(groupUuid, modifications);
```
SlimaneAmar, line 120: Call a method non transactional from a method transactional ?

SlimaneAmar (approved):

PR #695, opened 2025-09-25
No longer block the tomcat thread when waiting for import or export (servlet 3.0+ AsyncContext)

jonenst:
No longer block the tomcat thread when waiting for import or export (servlet 3.0+ AsyncContext)

Add no timeout configuration to behave like before this change

Spring forces us to adapt the tests with boilerplate code (asyncStarted() and asyncDispatch()) as per
https://docs.spring.io/spring-framework/reference/testing/mockmvc/hamcrest/async-requests.html
to mimic the servlet 3.0 redispatch

We avoid to store the return of the first perform in an MvcResult variable to avoid confusion (
The two mvc result represent the same request, inside they hold the same request/response
objects, but some other fields are different (asyncResult for example) and tests may
be unreliable if they use the first one (representing the state before the async split)
)

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/modifications/AsyncUtils.java
+++ b/src/main/java/org/gridsuite/modification/server/modifications/AsyncUtils.java
@@ -24,13 +24,16 @@ public final class AsyncUtils {
         // Should not be instantiated
     }
 
+    /**
+    * @param results should pass an empty list to be filled with the results
+    */
     // The signature of this method is chosen so that we can implement easily sequential or parallel schedule
     // If we change it (for example to parallel scheduling), we should keep the exceptional behavior consistent,
     // call the apply function inside a thenCompose anyway to wrap its exceptions in exceptional future completions.
-    public static CompletableFuture<List<Optional<NetworkModificationResult>>> scheduleApplyModifications(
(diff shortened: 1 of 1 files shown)


src/main/java/org/gridsuite/modification/server/modifications/NetworkModificationApplicator.java
```diff
@@ -108,13 +109,16 @@ public NetworkModificationResult applyModifications(ModificationApplicationGroup
...
-                .supplyAsync(() -> applyAndFlush(modificationInfosGroup, listener))
-                .join();
+                .supplyAsync(() -> applyAndFlush(modificationInfosGroup, listener));
         } else {
-            return applyAndFlush(modificationInfosGroup, listener);
+            return CompletableFuture.completedFuture(applyAndFlush(modificationInfosGroup, listener));
         }
     }
 
+    public NetworkModificationResult applyModificationsBlocking(ModificationApplicationGroup modificationInfosGroup, NetworkInfos networkInfos) {
```
SlimaneAmar: Move this code in a test class

jonenst: good catch thanks

PR #789, opened 2026-03-27
Add composite interactions

Meklo:
## PR Summary

Add a couple of composite operations and endpoints :
- Dedicated endpoint for moving composite submodifications 
- When retrieving excluded modifications parse through composites to check for excluded submodifications
- Propagate activation status from composite to its sub modifications except for description

Modify endpoint which retrieves composite submodifications metadata to keep track of each submodification's parent through a Map<UUID, List<ModificationInfos>> instead of a plain List<ModificationInfos> which looses track of this information 

### Why native queries are used for composite sub-modification reordering
Reordering sub-modifications within a composite requires manipulating the @OrderColumn on the join table. Hibernate implements this via UPDATE statements that cycle through intermediate position values, which temporarily assigns the same modification_id to two rows before the batch settles. This violates the unique constraint on modification_id at the statement level on both H2 (tests) and PostgreSQL (unless the constraint is deferred).
Rather than fighting the constraint, rewriteCompositeSubModifications bypasses Hibernate's collection management entirely using native DELETE + INSERT: it clears all join table rows for the composite, then reinserts them in the correct order. This is safe because it operates only on the join table, never touching the ModificationEntity rows themselves — so orphanRemoval and cascades are not triggered.


Mathieu-Deharbe (changes requested):
Bug when enabling all the nodes except one specific subnode : it is displayed as disabled but somehow it behaved as enabled during the apply : 
Edit : => actually I think it is being done in a separate ticket by @SlimaneAmar . I keep as a reference but it should be removed soon.

Image

Mathieu-Deharbe (commented):
Tests OK, a few discussions to be had.

souissimai (commented):
### some remarks

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -826,4 +860,113 @@ public List<ModificationInfos> insertCompositeModifications(
...
+    }
+
+    private void rewriteCompositeSubModifications(UUID compositeId, List<ModificationEntity> orderedSubMods) {
+        compositeModificationRepository.deleteCompositeSubModifications(compositeId);
+        for (int i = 0; i < orderedSubMods.size(); i++) {
+            compositeModificationRepository.insertCompositeSubModification(compositeId, orderedSubMods.get(i).getId(), i);
+        }
+    }
+
+    private boolean containsComposite(CompositeModificationEntity composite, UUID targetUuid) {
```
Mathieu-Deharbe: I think this would be better : 
```suggestion
    private boolean isInsideComposite(CompositeModificationEntity container, UUID targetUuid) {
```
Here the funciton name make it look like we are testing if the tested composite contain a composite.

Mathieu-Deharbe (approved):
Test and code OK. I am still not sure about `CompositeModificationRepository` but didn't find a better way and @SlimaneAmar says he is looking into it.

souissimai (approved):

SlimaneAmar (approved):

PR #804, opened 2026-04-23
use modifications_order for the composites

Mathieu-Deharbe:
## PR Summary
* replaces the OrderColumn annotation  used for the modifications contained inside a composite, with the modificationOrder ot ModificatoinEntity.
* removed removeOrphan because now modifications may be "orphan" when removed from composite and put into groups. They shoudn't be deleted in this case.
* few renaming and reorganisations along the way to make more readable what confused me (and remove some useless actions)


src/main/java/org/gridsuite/modification/server/entities/CompositeModificationEntity.java
```diff
@@ -32,12 +32,12 @@ public class CompositeModificationEntity extends ModificationEntity {
     @ColumnDefault("'My Composite'")
     private String name;
 
-    @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
+    @OneToMany(cascade = CascadeType.ALL)
```
Meklo, line 35: Could oprhanRemoval be kept as a security ?

Meklo (commented):
LGTM, just minor remark

src/main/java/org/gridsuite/modification/server/entities/CompositeModificationEntity.java
Mathieu-Deharbe: Since I do everything with entity lists without copying it like it was done before, JPA considers an entity to be orphan when I remove it from the list of movedEntities and automatically delete it. (while I just moved it to rootNode/group, but technically it is no longer referenced by any composite)
 
I could probably keep Orphan Removal by making a copy of the entity table, but it would lose a bit in performance and it would be less "JPA entity pure". 

I’m not sure what is best and if there is another way.

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -1002,29 +1013,20 @@ public void moveSubModification(@NonNull UUID groupUuid, UUID sourceCompositeUui
...
+            List<ModificationEntity> targetSubMods = targetComposite.getModifications();
+            insertModifications(targetSubMods, movedMods, beforeUuid);
         } else {
             // Moved to the root level of the network modification table
             ModificationGroupEntity group = getOrCreateModificationGroup(groupUuid);
             List<ModificationEntity> rootMods = group.getModifications().stream()
                     .filter(Objects::nonNull)
                     .filter(m -> !m.getStashed())
                     .collect(Collectors.toList());
-            movedEntities.forEach(movedEntity -> movedEntity.setGroup(group));
```
Meklo, line 982: I've noticed you've removed `movedEntities.forEach(movedEntity -> movedEntity.setGroup(group));` but kept `movedMods.forEach(entity -> entity.setGroup(null));` above line 979 is it deliberated ?

Mathieu-Deharbe: Well seen.
I did this because `movedEntities.forEach(movedEntity -> movedEntity.setGroup(group));` is already done in the following `group.setModifications(rootMods);` so it shouldn't be needed.

On the other hand when the group is set to null, we have to do it manually.

src/main/resources/db/changelog/changesets/changelog_20260423T141808Z.xml
```diff
@@ -0,0 +1,18 @@
...
+        <sqlFile
+                dbms="postgresql"
+                encoding="UTF-8"
+                path="migrationCompositesOrder.sql"
+                relativeToChangelogFile="true"
+                splitStatements="true"
+                stripComments="true"/>
+    </changeSet>
+    <changeSet author="deharbemat (generated)" id="1777454883407-19">
+        <dropPrimaryKey constraintName="composite_modification_sub_modificationsPK" tableName="COMPOSITE_MODIFICATION_SUB_MODIFICATIONS"/>
```
Meklo, line 13: Should we recreate a new primary key or modify the remaining unique constraint to also include id so that it acts as a PK ?

Mathieu-Deharbe: Good idea; I recreated it here : https://github.com/gridsuite/network-modification-server/pull/804/changes/5a6607fa6fb787decec390c5941f07ec1e16ee7f

(and merged the changesets into one)

Meklo (approved):

src/main/java/org/gridsuite/modification/server/repositories/CompositeModificationRepository.java
```diff
@@ -8,34 +8,12 @@
...
-    @Query(value = """
-        INSERT INTO composite_modification_sub_modifications (id, modification_id, modifications_order)
-        VALUES (:compositeId, :modificationId, :position)
-        """, nativeQuery = true)
-    void insertCompositeSubModification(
-            @Param("compositeId") UUID compositeId,
-            @Param("modificationId") UUID modificationId,
-            @Param("position") int position);
-}
+public interface CompositeModificationRepository extends JpaRepository<CompositeModificationEntity, UUID> { }
```
SlimaneAmar, line 19: Delete interface ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -188,12 +190,15 @@ public void updateCompositeModification(@NonNull UUID compositeUuid, @NonNull Li
...
+        for (int order = 0; order < copyEntities.size(); order++) {
+            copyEntities.get(order).setModificationsOrder(order);
+        }
         compositeEntity.getModifications().clear();
         compositeEntity.getModifications().addAll(copyEntities);
         modificationRepository.save(compositeEntity);
     }
 
-    private List<ModificationEntity> saveModificationsNonTransactional(UUID groupUuid, List<ModificationEntity> modifications) {
+    private List<ModificationEntity> addModificationsToGroupNonTransactional(@NonNull UUID groupUuid, List<ModificationEntity> modifications) {
```
SlimaneAmar: It's really a save !

src/main/java/org/gridsuite/modification/server/repositories/CompositeModificationRepository.java
Mathieu-Deharbe: Like we discussed earlier, I think we should keep it. It makes it possible to write this without any cast : 

`CompositeModificationEntity sourceComposite = compositeModificationRepository.findById(sourceCompositeUuid)`

Plus it is probably going to be used again soon here : https://github.com/gridsuite/network-modification-server/pull/812

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
Mathieu-Deharbe: Indeed but given that we manipulate entities the save is kind of implicit. I renamed the function because we send a list of modifications as a parameter and I think that it gives the impression that the list will be saved as the new list of the group. But it is added to the already present list and then all are saved.

Anyway this is not important, I reverted the renaming.

SlimaneAmar: Yes i  prefer the term `save`.
That's clearer, I think

SlimaneAmar (approved):

PR #806, opened 2026-04-28
Remove nesting limit check for composite

Meklo:


Mathieu-Deharbe (changes requested):
There are still parts that should be removed : 

In `createNetworkCompositeModification` : 

```
// Only composites contribute to depth — leaves have no entry in the join table and count as 0
        List<UUID> compositeUuids = new ArrayList<>(modificationRepository.findExistingCompositeModificationIds(modificationUuids));
        if (!compositeUuids.isEmpty()) {
            int maxDepth = getCompositesMaxDepthMap(compositeUuids).values().stream()
                    .mapToInt(Integer::intValue).max()
                    .orElse(0);
            if (maxDepth >= MAX_COMPOSITE_DEPTH) {
                throw new NetworkModificationException(COMPOSITE_NESTED_LIMIT_REACHED_ERROR);
            }
        }
```

including : 

`private static final int MAX_COMPOSITE_DEPTH = 5;`
and
`COMPOSITE_NESTED_LIMIT_REACHED_ERROR(HttpStatus.BAD_REQUEST),;`

Mathieu-Deharbe (approved):

PR #807, opened 2026-04-28
Add a modification reference as a new modification type

SlimaneAmar:


src/main/java/org/gridsuite/modification/server/entities/ModificationReferenceEntity.java
```diff
@@ -0,0 +1,80 @@
...
+@Entity
+@Table(name = "modification_reference")
+@PrimaryKeyJoinColumn(foreignKey = @ForeignKey(name = "modification_reference_id_fk_constraint"))
+public class ModificationReferenceEntity extends ModificationEntity {
+
+    @Column(name = "referenceId")
+    UUID referenceId;
+
+    @Column(name = "referenceType")
+    String referenceType;
```
Mathieu-Deharbe: Those will be turned into reference_id and reference_type in the database won't they ? If yes, why not directly : 

```suggestion
    @Column(name = "reference_id")
    UUID referenceId;

    @Column(name = "reference_type")
    String referenceType;
```
The reference would be more precise, easier to track.

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -480,6 +486,25 @@ private CompositeModificationInfos loadCompositeModificationMetadata(Modificatio
...
 
+    private ModificationReferenceInfos loadModificationReference(ModificationReferenceEntity modificationEntity) {
+        ModificationEntity referencedEntity = modificationRepository.getReferenceById(modificationEntity.getReferenceId());
+        if (referencedEntity == null) {
+            throw new NetworkModificationException(MODIFICATION_NOT_FOUND, String.format(MODIFICATION_NOT_FOUND_MESSAGE, modificationEntity.getReferenceId()));
+        }
+        return ModificationReferenceInfos.builder()
+            .uuid(modificationEntity.getId())
+            .messageType(modificationEntity.getMessageType())
+            .messageValues(modificationEntity.getMessageValues())
```
Mathieu-Deharbe: I am not sure of how the update will happen when the referenced modification is changed so I may be wrong. But just in case shouldn't we use the message type and message values of the `referencedEntity` here ?

Isn't it how we want this modification reference to be displayed in the front ?

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
```diff
@@ -783,6 +784,34 @@ void testCopyModificationWithUnexistingId() throws Exception {
...
+    @Test
+    void testCopyModificationReference() throws Exception {
+        // Create 1 modification reference
+        ModificationInfos loadModificationInfo = ModificationCreation.getCreationLoad("v1", "idLoad", "nameLoad", "1.1", LoadType.UNDEFINED);
+        loadModificationInfo = modificationRepository.saveModifications(UUID.randomUUID(), List.of(ModificationEntity.fromDTO(loadModificationInfo))).getFirst();
+        ModificationInfos modificationReferenceInfo = ModificationReferenceInfos.builder()
+            .referenceType(ModificationReferenceInfos.Type.SAMPLE)
+            .referenceId(loadModificationInfo.getUuid())
+            .referenceInfos(loadModificationInfo)
+            .stashed(false)
```
Mathieu-Deharbe, line 806: ? : 
```suggestion
```
(false by default)

Mathieu-Deharbe (commented):
Read but not executed yet. I don't want to mess with my database until this is completely finished.

src/main/java/org/gridsuite/modification/server/entities/ModificationReferenceEntity.java
```diff
@@ -0,0 +1,80 @@
...
+    @SneakyThrows
+    private void assignAttributes(ModificationReferenceInfos modificationReferenceInfos) {
+        modificationReferenceInfos.check();
+
+        this.referenceType = modificationReferenceInfos.getReferenceType().name();
+        this.referenceId = modificationReferenceInfos.getReferenceId();
+
+        // Appears as the referenced modification
+        this.setMessageType(modificationReferenceInfos.getMessageType());
+        this.setMessageValues(modificationReferenceInfos.getMessageValues());
```
Mathieu-Deharbe: Here it looks like you are just doing what is already done in ModificationEntity : 

```
private void assignAttributes(ModificationInfos modificationInfos) {
        this.setType(modificationInfos.getType().name());
        this.setMessageType(modificationInfos.getType().name());
        if (modificationInfos.getDescription() != null) {
            this.setDescription(modificationInfos.getDescription());
        }
        this.setMessageValues(new ObjectMapper().writeValueAsString(modificationInfos.getMapMessageValues()));
    }
```
And it crashes like this : 
`Error: TypeError: can't access property "equipmentId", modificationMetadata is null`
It looks like the data has not been initialised.

I think this would make more sense : 
```
        this.setMessageType(modificationReferenceInfos.getReferenceInfos().getMessageType());
        this.setMessageValues(new ObjectMapper().writeValueAsString(modificationReferenceInfos.getReferenceInfos().getMapMessageValues()));
```

Or do you want to use the messageType and messageValues from getReferenceInfos in order to display the modification reference just like what it references : ?

```suggestion
        // Appears as the referenced modification
        this.setMessageType(modificationReferenceInfos.getReferenceInfos().getType().name());
        this.setMessageValues(new ObjectMapper().writeValueAsString(modificationReferenceInfos.getReferenceInfos().getMapMessageValues()));
```
(I tested like that in my PR https://github.com/gridsuite/network-modification-server/pull/830/changes and it seems to be fine but this is kind of a )

src/main/java/org/gridsuite/modification/server/entities/ModificationReferenceEntity.java
SlimaneAmar: Removed

Mathieu-Deharbe (approved):

PR #809, opened 2026-04-29
Guarantee findAllChildrenUuids results order

Meklo:


src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -109,16 +109,15 @@ WHERE sm.id IN (?1)
...
-        "  UNION ALL " +
-        "  SELECT distinct m.modification_id" +
-        "  FROM composite_modification_sub_modifications m " +
-        "  INNER JOIN ModificationHierarchy mh ON m.id = mh.id " +
-        ") " +
-        "SELECT distinct cast(m.id AS VARCHAR) FROM ModificationHierarchy m ")
+    @NativeQuery("WITH RECURSIVE ModificationHierarchy (modification_id, path) AS ( " +
+            "  SELECT cast(:compositeUuid AS VARCHAR), ARRAY[0] " +
+            "  UNION ALL " +
+            "  SELECT cast(sm.modification_id AS VARCHAR), mh.path || (m.modifications_order + 1) " +
```
Mathieu-Deharbe: I don't think the "+ 1" is useful :
```suggestion
            "  SELECT cast(sm.modification_id AS VARCHAR), mh.path || m.modifications_order " +
```

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -109,16 +109,15 @@ WHERE sm.id IN (?1)
...
-        "  SELECT m0.id" +
-        "  FROM composite_modification_sub_modifications m0 " +
-        "  WHERE m0.id = :compositeUuid " +
-        "  UNION ALL " +
-        "  SELECT distinct m.modification_id" +
-        "  FROM composite_modification_sub_modifications m " +
-        "  INNER JOIN ModificationHierarchy mh ON m.id = mh.id " +
-        ") " +
-        "SELECT distinct cast(m.id AS VARCHAR) FROM ModificationHierarchy m ")
+    @NativeQuery("WITH RECURSIVE ModificationHierarchy (modification_id, path) AS ( " +
```
Mathieu-Deharbe, line 113: I would add a small comment like : 
`// returns the composite uuid and its children and subchildren's uuids`
because the request is a bit complicated and it is not very logical that compositeUuid is returned given the function name (`findAllChildrenUuids`).

Mathieu-Deharbe (approved):

PR #812, opened 2026-05-05
rename composite modification : update

souissimai:


src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -727,6 +732,14 @@ private void updateActivated(ModificationEntity entity, boolean activated) {
         }
     }
 
+    @SneakyThrows
+    private void renameCompositeModifications(CompositeModificationEntity compositeEntity, CompositeModificationInfos compositeMetadata) {
+        if (compositeMetadata.getName() != null) {
+            compositeEntity.setName(compositeMetadata.getName());
+            compositeEntity.setMessageValues(new ObjectMapper().writeValueAsString(compositeMetadata.getMapMessageValues()));
+        }
+    }
```
Mathieu-Deharbe: This could be transferred to `CompositeModificationRepository` (using `compositeModificationRepository`).

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -716,6 +718,9 @@ public void updateNetworkModificationMetadata(@NonNull List<UUID> modificationUu
             if (metadata.getActivated() != null) {
                 updateActivated(modificationEntity, metadata.getActivated());
             }
+            if (metadata instanceof CompositeModificationInfos compositeMetadata && modificationEntity instanceof CompositeModificationEntity composite) {
```
Mathieu-Deharbe: Given that the '!= null' are tested in this function for activated, description etc I would do : 
```suggestion
            if (metadata instanceof CompositeModificationInfos compositeMetadata
                    && modificationEntity instanceof CompositeModificationEntity composite
                    && composite.getName() != null) {
```
and remove the test from renameCompositeModifications which could take directly name as a string parameter.

souissimai: ok

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
souissimai: ok

Mathieu-Deharbe (approved):

PR #814, opened 2026-05-15
Make constraint deferrable to avoid error during move of composite

Meklo:


Mathieu-Deharbe (approved):

Mathieu-Deharbe (commented):
If I launch : 
`$ mvn clean compile liquibase:update liquibase:diff -Dpowsybl.liquibase.generate`

I get this in the resulting générated changelog : 
```
    <changeSet author="deharbemat (generated)" id="1779179167764-3">
        <dropPrimaryKey constraintName="COMPOSITE_MODIFICATION_SUB_MODIFICATIONSPK" tableName="COMPOSITE_MODIFICATION_SUB_MODIFICATIONS"/>
    </changeSet>
```

It looks like what you changed manually adds changelog noise data, probably because it doesn't correspond to the entity code.

It would be nice to clean it if possible.

PR #831, opened 2026-06-09
composite modifications action

souissimai:

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/repositories/CompositeModificationRepository.java
+++ b/src/main/java/org/gridsuite/modification/server/repositories/CompositeModificationRepository.java
@@ -11,9 +11,6 @@ import lombok.SneakyThrows;
 import org.gridsuite.modification.dto.CompositeModificationInfos;
 import org.gridsuite.modification.server.entities.CompositeModificationEntity;
 import org.springframework.data.jpa.repository.JpaRepository;
-import org.springframework.data.jpa.repository.Modifying;
-import org.springframework.data.jpa.repository.Query;
-import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
 
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -672,6 +685,26 @@ public void stashNetworkModifications(@NonNull List<UUID> modificationUuids, int
...
+            // delete the sub-modification before we can set it as stashed.
+            UUID parentCompositeId = modificationRepository.findCompositeIdByContainedModificationId(modificationUuid);
+            if (parentCompositeId != null) {
+                compositeModificationRepository.removeSubModification(parentCompositeId, modificationUuid);
+                ModificationGroupEntity parentGroup = compositeModificationRepository.findById(parentCompositeId)
+                        .orElseThrow(() -> new NetworkModificationException(MODIFICATION_NOT_FOUND))
+                        .getGroup();
+                modificationEntity.setGroup(parentGroup);
+                modificationEntity.setStashed(false);
+            }
```
Meklo: Could be handled directly from NetworkModificationService::stashNetworkModifications as a move from the composite to the groupUuid : 

        for (UUID modificationUuid : modificationUuids) {
            UUID parentCompositeUuid = modificationRepository.findCompositeIdByContainedModificationId(modificationUuid);
            if (parentCompositeUuid != null) {
                networkModificationRepository.moveSubModification(groupUuid, parentCompositeUuid, null, modificationUuid, null);
            }
        }
        networkModificationRepository.stashNetworkModifications(modificationUuids,
                networkModificationRepository.getModificationsCount(groupUuid, true));

So that you don't have to deal with the remove oprhanRemoval issue, what do you think ?

src/main/java/org/gridsuite/modification/server/repositories/CompositeModificationRepository.java
```diff
@@ -26,4 +29,22 @@ default void renameCompositeModifications(CompositeModificationEntity compositeE
...
+     * entity intact so it can be reassigned to the parent group and stashed normally.
+     */
+    @Modifying
+    @Query(value = """
+            DELETE FROM composite_modification_sub_modifications
+            WHERE id = :compositeUuid
+              AND modification_id = :subModificationUuid
+            """, nativeQuery = true)
+    void removeSubModification(@Param("compositeUuid") UUID compositeUuid,
+                               @Param("subModificationUuid") UUID subModificationUuid);
```
Meklo: Could be removed if the submodification stashing is handled as a move before the stash

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
```diff
@@ -383,14 +383,32 @@ public void deleteNetworkModifications(UUID groupUuid, List<UUID> modificationsU
...
+        List<UUID> subModificationUuids = modificationsToMoveUuids.stream()
+                .filter(uuid -> !originRootUuids.contains(uuid))
+                .filter(uuid -> !hasSelectedAncestor(uuid, selection, parentByChild))
+                .toList();
+        for (UUID uuid : subModificationUuids) {
+            UUID parentCompositeUuid = parentByChild.get(uuid);
+            if (parentCompositeUuid != null) {
+                networkModificationRepository.moveSubModification(originGroupUuid, parentCompositeUuid, null, uuid, null);
+            }
+        }
```
Meklo: This part could get quite costly if we systematically retrieve all the modifications of a group for each move operation.
How about directly retrieving the children of composites contained in modificationsToMoveUuids through findExistingCompositeModificationIds and findAllChildrenUuids? Then, iterating through modificationsToMoveUuids, you can skip those that will be moved with their ancestor, and directly retrieve the parent composite of the remaining subModifications with findCompositeIdByContainedModificationId.
That way you can also avoid the ancestor map buildup

Meklo (changes requested):

Meklo (approved):

PR #838, opened 2026-06-19
Reference modification

souissimai:


src/main/java/org/gridsuite/modification/server/entities/CompositeModificationEntity.java
```diff
@@ -47,7 +49,10 @@ public CompositeModificationEntity(@NonNull CompositeModificationInfos composite
 
     @Override
     public CompositeModificationInfos toModificationInfos() {
-        List<ModificationInfos> modificationsInfos = modifications.stream().map(ModificationEntity::toModificationInfos).toList();
+        List<ModificationInfos> modificationsInfos = modifications.stream()
+                .map(ModificationEntity::toModificationInfos)
+                .peek(this::fillDisplayMessage)
```
Mathieu-Deharbe: peek is a bit dangerous to use. From sonar rules : 

> A key difference with other intermediate Stream operations is that the Stream implementation is free to skip calls to peek() for optimization purpose. This can lead to peek() being unexpectedly called only for some or none of the elements in the Stream.

Maybe use foreach instead after the creation of the array ? Or some other stream funciton ?

src/test/java/org/gridsuite/modification/server/CompositeControllerTest.java
```diff
@@ -184,7 +184,7 @@ void testSplit() throws Exception {
                         .contentType(MediaType.APPLICATION_JSON),
                 status().isOk());
 
-        assertApplicationStatusOK(mvcResult);
+        //assertApplicationStatusOK(mvcResult);
```
Mathieu-Deharbe: Forgotten ? I uncommented it and the test passed.
```suggestion
        assertApplicationStatusOK(mvcResult);
```

src/test/java/org/gridsuite/modification/server/service/ModificationIndexationTest.java
```diff
@@ -309,9 +309,10 @@ void testInsertCompositeModifications() {
         Split this composite and insert the contained modifications to group 2, variant 2
          */
         UUID groupUuid2 = UUID.randomUUID();
+
         Pair<List<CompositeInfos>, List<ModificationApplicationContext>> modificationContextInfos = Pair.of(
                 List.of(new CompositeInfos(compositeUuid, "", false)),
-                List.of(new ModificationApplicationContext(networkInfos.getNetworkUuuid(), variant2, UUID.randomUUID(), UUID.randomUUID()))
+                 List.of(new ModificationApplicationContext(networkInfos.getNetworkUuuid(), variant2, UUID.randomUUID(), UUID.randomUUID()))
```
Mathieu-Deharbe:
```suggestion
                List.of(new ModificationApplicationContext(networkInfos.getNetworkUuuid(), variant2, UUID.randomUUID(), UUID.randomUUID()))
```

PR #844, opened 2026-07-03
Fix composite assembling reordering

Meklo:

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
+++ b/src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
@@ -988,11 +988,8 @@ public class NetworkModificationRepository {
                 .findFirst()
                 .orElse(null);
         if (originGroup != null) {
-            List<ModificationEntity> originGroupModifications = new ArrayList<>(originGroup.getModifications());
+            List<ModificationEntity> originGroupModifications = originGroup.getModifications();
             originGroupModifications.removeIf(mod -> assembledModificationsUuids.contains(mod.getId()));
-            for (int i = 0; i < originGroupModifications.size(); i++) {
-                originGroupModifications.get(i).setModificationsOrder(i);
-            }
             originGroup.setModifications(originGroupModifications);
             assembledModifications.forEach(modificationEntity -> modificationEntity.setGroup(null));
         }
@@ -1002,13 +999,10 @@ public class NetworkModificationRepository {
             if (compositeUuid != null) {
                 CompositeModificationEntity previousOwner = compositeModificationRepository.findById(compositeUuid).orElse(null);
                 if (previousOwner != null) {
-                    List<ModificationEntity> modificationsLeft = new ArrayList<>(previousOwner.getModifications()
+                    List<ModificationEntity> modificationsLeft = previousOwner.getModifications()
                             .stream()
                             .filter(mod -> !assembledModificationsUuids.contains(mod.getId()))
-                            .toList());
-                    for (int i = 0; i < modificationsLeft.size(); i++) {
-                        modificationsLeft.get(i).setModificationsOrder(i);
-                    }
+                            .toList();
                     previousOwner.setModifications(modificationsLeft);
                 }
             }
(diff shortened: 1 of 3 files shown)


src/test/java/org/gridsuite/modification/server/CompositeControllerTest.java
Mathieu-Deharbe: This is a new test : 
```suggestion
        assertEquals(1, networkModificationRepository.getModifications(TEST_GROUP_ID, true, true).size());
```

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -1016,15 +1022,19 @@ public CompositeModificationEntity assembleNetworkModificationsIntoNewComposite(
...
             targetGroup.setModifications(modifications);
         } else if (targetComposite != null) {
-            List<ModificationEntity> modifications = targetComposite.getModifications();
+            List<ModificationEntity> modifications = new ArrayList<>(targetComposite.getModifications());
             modifications.add(targetIndex, newCompositeEntity);
-            for (int i = 0; i < targetComposite.getModifications().size(); i++) {
-                targetComposite.getModifications().get(i).setModificationsOrder(i);
+            for (int i = 0; i < modifications.size(); i++) {
+                modifications.get(i).setModificationsOrder(i);
             }
```
Mathieu-Deharbe: This reordering is already done in `targetComposite.setModifications(modifications);` just after.

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -984,8 +984,11 @@ public CompositeModificationEntity assembleNetworkModificationsIntoNewComposite(
                 .findFirst()
                 .orElse(null);
         if (originGroup != null) {
-            List<ModificationEntity> originGroupModifications = originGroup.getModifications();
+            List<ModificationEntity> originGroupModifications = new ArrayList<>(originGroup.getModifications());
             originGroupModifications.removeIf(mod -> assembledModificationsUuids.contains(mod.getId()));
+            for (int i = 0; i < originGroupModifications.size(); i++) {
+                originGroupModifications.get(i).setModificationsOrder(i);
+            }
```
Mathieu-Deharbe: I think this reordering should be done in `originGroup.setModifications(originGroupModifications);` just after. It would be similar to how reordering works in composites' `setModifications` and would probably be enough to correct the bug.

I added : 
```
for (int i = 0; i < this.modifications.size(); i++) {
            this.modifications.get(i).setModificationsOrder(i);
        }
```
in `setModifications` and didn't manage to reproduce the bug.

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -1016,15 +1022,19 @@ public CompositeModificationEntity assembleNetworkModificationsIntoNewComposite(
...
-            List<ModificationEntity> modifications = targetGroup.getModifications();
+            List<ModificationEntity> modifications = new ArrayList<>(targetGroup.getModifications());
             modifications.add(targetIndex, newCompositeEntity);
+            for (int i = 0; i < modifications.size(); i++) {
+                modifications.get(i).setModificationsOrder(i);
+            }
             targetGroup.setModifications(modifications);
         } else if (targetComposite != null) {
-            List<ModificationEntity> modifications = targetComposite.getModifications();
+            List<ModificationEntity> modifications = new ArrayList<>(targetComposite.getModifications());
```
Mathieu-Deharbe: I don't see the point of creating of a new array. We are in a Transactional function, why don't we directly edit the entities arrays ? Persistance should do the job.
If I am right here the reordering would be enough, no need to reapply setModifications.

Mathieu-Deharbe (changes requested):
I think it can be done simply by adding reordering in the `setModifications` of `ModificationGroupEntity` but I am not sure. It seems to work but I think there was a reason why the reordering is not already there.

Edit : OK I am definitely not sure : I think I reproduced it, I will look more into it.

src/test/java/org/gridsuite/modification/server/CompositeControllerTest.java
Meklo: I think I only swapped modificationRepository for networkModificationRepository here ? I'm not sure I get it

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
Meklo: I may have overblown the fix haha

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
Meklo: I initially figured the same but wasn't able to save the new order without making a copy first in the end, I'll double check

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
Meklo: That would be ideal yes I'll try it out

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
Meklo: I now get it. I added multiple calls to CompositeModificationEntity::setModifications which does a this.modifications.clear();. If we don't send an hard copy of modifications before calling it on itself it won't work. I rolled back changes to assembleNetworkModificationsIntoNewComposite and added reordering to ModificationGroupEntity setModifications which seems to work just fine

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
Meklo: It seems to do the trick just fine !

Mathieu-Deharbe (approved):
I tried very hard to reproduce the bug and it seems fine.

Maybe we should warn Thibault that the node where he used that function may have some ordering problems left. Better to try from a new node.

I am still a bit worried because I think some time ago I almost added that reordering here, but didn't do it because of a good reason. But I can't remember... We will see.

PR #849, opened 2026-07-15
fix : ignore stashed when reordering groups

Mathieu-Deharbe:
ignore stashed when reordering groups

Plus small refacto in order to use the same `getNotStashedModifications` new function several time in the code.

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/entities/ModificationGroupEntity.java
+++ b/src/main/java/org/gridsuite/modification/server/entities/ModificationGroupEntity.java
@@ -11,6 +11,7 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/gridsuite/modification/server/entities/ModificationGroupEntity.java
```diff
@@ -49,11 +50,20 @@ public void addModification(ModificationEntity modification) {
         }
     }
 
+    /**
+     * @return a mutable ArrayList of the modifications without those stashed
+     */
+    public List<ModificationEntity> getNotStashedModifications() {
```
Meklo: What do you think of getActiveModifications ? NotStashed sounds a bit weird to me

Meklo (approved):
Works as intended! Just a minor remark, do as you see fit

src/main/java/org/gridsuite/modification/server/entities/ModificationGroupEntity.java
Mathieu-Deharbe: Done : https://github.com/gridsuite/network-modification-server/pull/849/changes/2013186bd8c18c944b0422844f0dd0a5b3bc072b

PR #850, opened 2026-07-16
add voltage regulation for battery creation

EtienneLt:


src/main/resources/db/changelog/changesets/changelog_20260716T122908Z.xml
```diff
@@ -0,0 +1,28 @@
...
+    <changeSet author="lesoteti (generated)" id="1784204970270-5">
+        <addColumn tableName="battery_creation">
+            <column name="targetv" type="FLOAT(53)"/>
+        </addColumn>
+    </changeSet>
+    <changeSet author="lesoteti (generated)" id="1784204970270-6">
+        <addColumn tableName="battery_creation">
+            <column name="voltage_regulation_on" type="BOOLEAN"/>
+        </addColumn>
+    </changeSet>
```
Mathieu-Deharbe, line 27: Maybe you should set dfault values, or a migration. Or improve the handling of voltageRegulationOn null values, because the loading of my previously created battery creation failed because of a lack of data : 

Image

Edit : actually I think you should turn `private boolean voltageRegulationOn;` into a Boolean (nullable) in order to fix the problem. (in the entity and in the ModificationInfos)

Mathieu-Deharbe (approved):

Mathieu-Deharbe (changes requested):
Misclick, sorry : the null handling has to be added.

Mathieu-Deharbe (approved):

PR #852, opened 2026-07-19
referenced modifications

souissimai:


src/main/java/org/gridsuite/modification/server/repositories/CompositeModificationRepository.java
```diff
@@ -32,4 +33,14 @@ default void renameCompositeModification(CompositeModificationEntity compositeEn
         compositeEntity.setName(name);
         compositeEntity.setMessageValues(new ObjectMapper().writeValueAsString(compositeEntity.toModificationInfos().getMapMessageValues()));
     }
+
+    @SneakyThrows
+    default void loadCompositeModification(ModificationInfos modificationInfos) {
```
Mathieu-Deharbe: Actually this function doesn't load a composite modification. It only generates the messages, and for any kind of modification.
I think it should be in a more generic repository.
```suggestion
    default void generateModificationMessage(ModificationInfos modificationInfos) {
```

src/test/java/org/gridsuite/modification/server/modifications/ModificationReferenceTest.java
```diff
@@ -78,6 +84,97 @@ protected void testUpdateModificationMessage(ModificationInfos modificationInfos
...
+        assertNotNull(resultChild.getMessageValues());
+    }
+
+    @Test
+    void testCompositeToModificationInfosPreservesExistingDisplayMessage() {
+        // child already carries a messageType/messageValues: withDisplayMessage() must
+        // leave them untouched (the "non-null" branch of each if).
+        ModificationInfos child = ModificationCreation.getCreationLoad("v1", "idLoad", "nameLoad", "1.1", LoadType.UNDEFINED);
+        child.setMessageType("LOAD_CREATION");
+        child.setMessageValues("{\"equipmentId\":\"idLoad\"}");
```
Mathieu-Deharbe: This test doesn't seem to work. If I change this like this : 
```suggestion
    void testCompositeToModificationInfosPreservesExistingDisplayMessage() {
        // child already carries a messageType/messageValues: withDisplayMessage() must
        // leave them untouched (the "non-null" branch of each if).
        ModificationInfos child = ModificationCreation.getCreationLoad("v1", "idLoad", "nameLoad", "1.1", LoadType.UNDEFINED);
        child.setMessageType("LOAD_CREATION_dummy");
        child.setMessageValues("{\"equipmentId\":\"idLoad_dummy\"}");
```
The test still succeed.s So it looks like those are not preserved.

BTW I agree with coderabbit : the messageType and messageValues added at the beginning are identical to what would be generated by the code. So it doesn't really test if they are preserved.

src/test/java/org/gridsuite/modification/server/modifications/ModificationReferenceTest.java
```diff
@@ -78,6 +84,97 @@ protected void testUpdateModificationMessage(ModificationInfos modificationInfos
...
+    }
+
+    @Test
+    void testGetModificationReferenceInfoPreservesExistingChildDisplayMessageFromRepository() {
+        // Same DB-reload path as above, but the child already has custom messageType/messageValues:
+        // the null-guards in loadModificationReference() must leave them untouched, not overwrite
+        // them with the computed fallback. Covers the false branch of the two new ifs.
+        ModificationInfos child = ModificationCreation.getCreationLoad("v1", "idLoad", "nameLoad", "1.1", LoadType.UNDEFINED);
+        child.setMessageType(ModificationType.LOAD_CREATION.name());
+        child.setMessageValues("{\"equipmentId\":\"idLoad\"}");
```
Mathieu-Deharbe: Same remark than above. If I just do this : 
```suggestion
        child.setMessageType(ModificationType.LOAD_MODIFICATION.name());
        child.setMessageValues("{\"equipmentId\":\"idLoad_dummy\"}");
```
the test still succeeds.

So the test seem inconsistent with its tests name (and comment).

src/test/java/org/gridsuite/modification/server/modifications/ModificationReferenceTest.java
```diff
@@ -88,4 +185,43 @@ private ModificationInfos buildCompositeModification() {
             .stashed(false)
             .build();
     }
+
+    @Test
+    void testGetModificationReferenceToNonComposite() {
```
Mathieu-Deharbe, line 135: It is good to test it because technically it can be done but I think we should discuss with Slimane and Thibault if there really is a chance for this to exist in gridsuite.

If not, maybe we should rigidify all this and simply assume that any referenceInfos if a composoite data.

src/test/java/org/gridsuite/modification/server/modifications/ModificationReferenceTest.java
```diff
@@ -88,4 +185,43 @@ private ModificationInfos buildCompositeModification() {
...
+        List<ModificationInfos> saved = networkModificationRepository.saveModificationInfos(UUID.randomUUID(), List.of(referenceInfos));
+
+        ModificationInfos fetched = networkModificationRepository.getModificationInfo(saved.get(0).getUuid());
+
+        assertInstanceOf(ModificationReferenceInfos.class, fetched);
+        assertNotNull(((ModificationReferenceInfos) fetched).getReferenceInfos());
+    }
+
+    @Test
+    void testFillDisplayMessageKeepsExistingValues() {
```
Mathieu-Deharbe: The test name doesn't have much to do with the code. Maybe simply ? : 
```suggestion
    void testFillDisplayMessage() {
```

Mathieu-Deharbe (changes requested):

src/test/java/org/gridsuite/modification/server/modifications/ModificationReferenceTest.java
souissimai: This could be addressed as a future enhancement. to be checked!

src/test/java/org/gridsuite/modification/server/modifications/ModificationReferenceTest.java
souissimai: right

src/test/java/org/gridsuite/modification/server/modifications/ModificationReferenceTest.java
souissimai: ok

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -502,7 +502,12 @@ private ModificationInfos loadModificationReference(ModificationEntity modificat
             ModificationEntity referencedEntity = modificationRepository.findAllByIdIn(List.of(referenceEntity.getReferenceId())).stream().findFirst()
                 .orElseThrow(() -> new NetworkModificationException(MODIFICATION_NOT_FOUND, String.format(MODIFICATION_NOT_FOUND_MESSAGE, referenceEntity.getReferenceId())));
             ModificationReferenceInfos modificationReferenceInfos = referenceEntity.toModificationInfos();
-            modificationReferenceInfos.setReferenceInfos(toModificationsInfosOptimized(referencedEntity));
+            ModificationInfos refInfos = toModificationsInfosOptimized(referencedEntity);
+
+            if (refInfos instanceof CompositeModificationInfos composite && composite.getModificationsInfos() != null) {
+                composite.getModificationsInfos().forEach(compositeModificationRepository::generateModificationMessage);
+            }
+            modificationReferenceInfos.setReferenceInfos(refInfos);
```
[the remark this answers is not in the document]
souissimai: fixed

src/main/java/org/gridsuite/modification/server/repositories/CompositeModificationRepository.java
```diff
@@ -32,4 +33,14 @@ default void renameCompositeModification(CompositeModificationEntity compositeEn
...
+
+    @SneakyThrows
+    default void generateModificationMessage(ModificationInfos modificationInfos) {
+        if (modificationInfos.getMessageType() == null) {
+            modificationInfos.setMessageType(modificationInfos.getType().name());
+        }
+        if (modificationInfos.getMessageValues() == null) {
+            modificationInfos.setMessageValues(new ObjectMapper().writeValueAsString(modificationInfos.getMapMessageValues()));
+        }
+    }
```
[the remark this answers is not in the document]
souissimai: done

Mathieu-Deharbe (approved):

PR #853, opened 2026-07-20
refactor: simplify exception handling

GhilesA:
## PR Summary

use new networkmodification lib exception as base error and default ones in other cases


src/main/java/org/gridsuite/modification/server/error/NetworkModificationExceptionHandler.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+
+/**
+ * @author Slimane Amar <slimane.amar at rte-france.com>
+ */
+@ControllerAdvice
+public class NetworkModificationExceptionHandler {
```
Meklo: Should extend powsybl commons AbstractBusinessExceptionHandler in order to make it fit the system like in other servers, see https://github.com/powsybl/powsybl-ws-commons/blob/a63cedca470bded8553be0bf3cfe660130a1f149/src/main/java/com/powsybl/ws/commons/error/AbstractBusinessExceptionHandler.java

src/main/java/org/gridsuite/modification/server/entities/ModificationEntity.java
```diff
@@ -141,6 +141,9 @@ public static ModificationEntity fromDTO(ModificationInfos dto) {
                 Constructor<? extends ModificationEntity> constructor = entityClass.getConstructor(dto.getClass());
                 return constructor.newInstance(dto);
             } catch (Exception e) {
+                if (e.getCause() instanceof NetworkModificationException networkModificationException) {
```
Meklo: Multiple catch clauses wouldn't do the trick here ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -334,7 +334,7 @@ public List<ModificationInfos> getModifications(UUID groupUuid, boolean onlyMeta
         try {
             return onlyMetadata ? getModificationsMetadata(groupUuid, onlyStashed) : getModificationsInfos(List.of(groupUuid), onlyStashed);
         } catch (NetworkModificationException e) {
-            if (e.getType() == MODIFICATION_GROUP_NOT_FOUND && !errorOnGroupNotFound) {
+            if (e.getMessage().startsWith(MODIFICATION_GROUP_NOT_FOUND.getMessage()) && !errorOnGroupNotFound) {
```
Meklo: I kinda agree with coderabbit here, wouldn't type comparison work here ?

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
```diff
@@ -1635,12 +1645,12 @@ void testVerifyModifications() throws Exception {
         // try to verify unexisting modification
         mockMvc.perform(get("/v1/groups/{groupId}/network-modifications/verify", TEST_GROUP_ID)
             .param("uuids", UUID.randomUUID().toString()))
-            .andExpect(status().isNotFound());
```
Meklo, line 1642: errors which initially returned a 404 should still return it instead of a 500 imo

PR #867, opened 2026-08-07
Store the applicability per root network tag on modifications

flomillot:
Depends on gridsuite/network-modification#219.

The applicability per root network was held by study-server, attached to a (node, root network) pair, so it was lost as soon as a modification left its node — export to GridExplore then re-import, or copy/paste between studies. It is now carried by the modification itself, keyed by root network tag.

- `ModificationEntity` holds `applicabilityByRootNetworkTag` in the `modification_root_network_applicability` table, copied into every DTO next to `activated`, so it follows the modification through every duplication path.
- `PUT /v1/network-modifications/root-network-applicability` updates it for a tag, propagating to the sub modifications of a composite like the global flag does, and reading it needs no dedicated endpoint: it travels with the modifications, the metadata ones included, where it is read for the whole batch in a single query and resolved to the shared modification when the entry is a reference.
- The application context and `BuildInfos` now carry the root network tag instead of a list of excluded modification uuids, and the applicator filters on `isActivatedOn(tag)`, the single predicate shared by the five call sites.
- A reference to a shared modification does not own its applicability: it uses the one of the modification it points to.

Behaviour change: `isActivatedOn` now covers the stashed flag too, so posting a modification with `stashed=true` no longer applies it to the network. It is stored stashed, as before.


src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -291,6 +291,26 @@ public ResponseEntity<Void> updateNetworkModificationMetadata(
...
     }
 
+    @PutMapping(value = "/network-modifications/root-network-applicability")
+    @Operation(summary = "Updates the applicability of network modifications for a root network tag")
+    @ApiResponse(responseCode = "200", description = "The applicability of the network modifications has been successfully updated")
+    public ResponseEntity<Void> updateRootNetworkApplicability(
+            @Parameter(description = "Network modifications UUIDs") @RequestParam("uuids") List<UUID> networkModificationUuids,
+            @Parameter(description = "Root network tag") @RequestParam("rootNetworkTag") String rootNetworkTag,
+            @Parameter(description = "Applicability on this root network tag") @RequestParam("activated") boolean activated) {
+        networkModificationService.updateRootNetworkApplicability(networkModificationUuids, rootNetworkTag, activated);
```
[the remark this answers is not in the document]
flomillot: fixed

src/main/resources/db/changelog/changesets/changelog_20260806T112115Z.xml
```diff
@@ -0,0 +1,17 @@
...
+                <constraints nullable="false" primaryKey="true" primaryKeyName="modification_root_network_applicabilityPK"/>
+            </column>
+            <column name="applicable" type="BOOLEAN"/>
+            <column name="root_network_tag" type="VARCHAR(4)">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="modification_root_network_applicabilityPK"/>
+            </column>
+        </createTable>
+    </changeSet>
+    <changeSet author="florent (generated)" id="1786015296336-8">
+        <addForeignKeyConstraint baseColumnNames="modification_id" baseTableName="modification_root_network_applicability" constraintName="modification_root_network_applicability_fk" deferrable="false" initiallyDeferred="false" referencedColumnNames="id" referencedTableName="modification" validate="true"/>
```
Mathieu-Deharbe, line 15: It seems to me that if `deferrable="false"` then `initiallyDeferred="false"` is not useful.

Mathieu-Deharbe (commented):
.applicabilityByRootNetworkTag(copyApplicabilityByRootNetworkTag()) is missing for 

`public ModificationReferenceInfos toModificationInfos() {`

I am not sure that this will be useful though, depending of how the applicability is handled for reference modifications.

**Edit :** ok nevermind it looks like it is well handled and tested in the code.

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -41,8 +41,9 @@ public interface ModificationRepository extends JpaRepository<ModificationEntity
     @Query(value = "SELECT m FROM ModificationEntity m WHERE m.container.id = ?1 AND m.stashed = ?2 order by m.modificationsOrder")
     List<ModificationEntity> findAllByContainerId(@Param("containerId") UUID containerId, @Param("stashed") Boolean stashed);
 
-    @Query(value = "SELECT m FROM ModificationEntity m WHERE m.container.id = ?1 AND m.stashed = false AND m.activated = true AND m.id NOT IN (?2) order by m.modificationsOrder")
-    List<ModificationEntity> findAllActiveModificationsByContainerId(UUID containerId, Set<UUID> excludedList);
+    // the applicability per root network tag is not resolved here: it is filtered when applying the modifications
+    @Query(value = "SELECT m FROM ModificationEntity m WHERE m.container.id = ?1 AND m.stashed = false AND m.activated = true order by m.modificationsOrder")
+    List<ModificationEntity> findAllActiveModificationsByContainerId(UUID containerId);
```
Mathieu-Deharbe: Ok it works but wouldn't it be better to keep that root network tag filtering here, even if it complicates the request ?

Without it many useless ModificationEntity will be loaded and then not applied.

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -830,6 +842,47 @@ private void updateActivated(ModificationEntity entity, boolean activated) {
...
+    public void updateRootNetworkApplicability(@NonNull List<UUID> modificationUuids, @NonNull String rootNetworkTag, boolean applicable) {
+        modificationUuids.forEach(modificationUuid -> updateRootNetworkApplicability(getModificationEntity(modificationUuid), rootNetworkTag, applicable));
+    }
+
+    private void updateRootNetworkApplicability(ModificationEntity entity, String rootNetworkTag, boolean applicable) {
+        ModificationEntity applicabilityHolder = getApplicabilityHolder(entity);
+        applicabilityHolder.getApplicabilityByRootNetworkTag().put(rootNetworkTag, applicable);
+        if (applicabilityHolder instanceof CompositeModificationEntity composite) {
+            composite.getModifications().forEach(sub -> updateRootNetworkApplicability(sub, rootNetworkTag, applicable));
+        }
```
Mathieu-Deharbe, line 916: This was useful when clicking on a root network tag icon of a composite had to disable also its content but I am quite sure that this is not the final desired behavior.

Now that the composite hierarchy handles the application well (if the composite has a root network tag disabled, its content is not applied) maybe this part should be removed.

But to be checked with the PO of course.

(I checked, it works fine, only the ux would have to be improved in a separate ticket)

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -291,6 +291,26 @@ public ResponseEntity<Void> updateNetworkModificationMetadata(
...
+            @Parameter(description = "Root network tag") @RequestParam("rootNetworkTag") String rootNetworkTag,
+            @Parameter(description = "Applicability on this root network tag") @RequestParam("applicable") boolean applicable) {
+        networkModificationService.updateRootNetworkApplicability(networkModificationUuids, rootNetworkTag, applicable);
+        return ResponseEntity.ok().build();
+    }
+
+    @GetMapping(value = "/groups/{groupUuid}/root-network-applicability", produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(summary = "Get the applicability per root network tag of all the network modifications in a group, sub modifications included")
+    @ApiResponse(responseCode = "200", description = "The applicabilities were returned")
+    public ResponseEntity<Map<UUID, Map<String, Boolean>>> getRootNetworkApplicabilities(
```
Mathieu-Deharbe: probably obsolete, cf : https://github.com/gridsuite/study-server/pull/1053#discussion_r3821287171

src/test/java/org/gridsuite/modification/server/modifications/tabularcreations/TabularGeneratorCreationsTest.java
```diff
@@ -179,7 +179,7 @@ void testCheckSqlRequestsCount() throws Exception {
                         status().isOk(), content().contentType(MediaType.APPLICATION_JSON))
                 .andReturn();
         // We check that the request count is not dependent on the number of sub creations of the tabular creation (the JPA N+1 problem is correctly solved)
-        assertSelectCount(5);
+        assertSelectCount(6);
```
Mathieu-Deharbe, line 182: From what I have seen the extra request is : 

```
    select
        abrnt1_0.modification_id,
        abrnt1_0.root_network_tag,
        abrnt1_0.applicable 
    from
        modification_root_network_applicability abrnt1_0 
    where
        abrnt1_0.modification_id in
```

It makes sense but given how modification request are always heavily optimised, we might have to merge this into the main request (along with activated).

src/test/java/org/gridsuite/modification/server/modifications/tabularmodifications/TabularGeneratorModificationsTest.java
```diff
@@ -132,7 +132,7 @@ void testSqlRequestsCountOnGetGroupModifications() throws Exception {
 
         reset();
         List<ModificationInfos> tabularModifications = ApiUtils.getGroupModifications(mockMvc, getGroupId()); // Getting two tabular modifications with respectively one and three sub-modifications
-        assertSelectCount(10); // 10 before improvements
+        assertSelectCount(12); // 10 before improvements
```
Mathieu-Deharbe: The comments are obsoletes, including in some other files about request counts.

src/test/java/org/gridsuite/modification/server/service/ModificationRepositoryTest.java
```diff
@@ -1519,4 +1522,79 @@ void testCreateCouplingDevice() {
...
+    }
+
+    @Test
+    void testUpdateRootNetworkApplicabilityPropagatesToCompositeSubModifications() {
+        UUID compositeUuid = insertComposite(TEST_GROUP_ID_2, false, "v1d1", "v1d2");
+
+        networkModificationRepository.updateRootNetworkApplicability(List.of(compositeUuid), ROOT_NETWORK_TAG, false);
+
+        Map<UUID, Map<String, Boolean>> applicabilities = networkModificationRepository.getRootNetworkApplicabilities(TEST_GROUP_ID_2);
+        assertEquals(3, applicabilities.size(), "The composite and its two sub modifications are all reported");
```
Mathieu-Deharbe: I don't think reported is the right word. Did you mean ? : 

```suggestion
        assertEquals(3, applicabilities.size(), "The composite and its two sub modifications applicabilities must all be updated");
```

src/main/resources/db/changelog/changesets/changelog_20260806T112115Z.xml
flomillot: Ah okay, I just used the liquid base command. Should I change it then ?

flomillot: All files contain s this pattern, so I don't think I should change it only here.

Mathieu-Deharbe: No, don't change it. 
Even if it is not useful, it won't cause any problem so let's trust liquidbase.

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
flomillot: I was not aware of all the SQL optimization about fetching modification, but you're right, so I updated it

src/test/java/org/gridsuite/modification/server/modifications/tabularcreations/TabularGeneratorCreationsTest.java
flomillot: If you mean when we request modification metadata with the latest code, I don't think we can do better.

src/test/java/org/gridsuite/modification/server/modifications/tabularmodifications/TabularGeneratorModificationsTest.java
flomillot: Yeah I just removed these comments in the lines I modified.

src/test/java/org/gridsuite/modification/server/service/ModificationRepositoryTest.java
flomillot: Removed

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
flomillot: Yes good point, let's wait PO decision

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
```diff
@@ -383,23 +406,21 @@ public NetworkModificationResult buildVariant(@NonNull UUID networkUuid, @NonNul
...
-                modificationGroupsInfos.add(new ModificationApplicationGroup(groupUuid, modifications, reportInfos));
+                modificationGroupsInfos.add(new ModificationApplicationGroup(groupUuid, modifications, reportInfos, buildInfos.getRootNetworkTag()));
 
             }
         );
 
         PreloadingStrategy preloadingStrategy = modificationGroupsInfos.stream().map(ModificationApplicationGroup::modifications)
             .flatMap(Collection::stream)
-            .filter(m -> m.getActivated() && !m.getStashed())
             .map(ModificationInfos::getType)
```
Meklo, line 424: Preloading strategy was defined by active modifications, maybe the filter should stay and use isActivatedOn no ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -865,6 +903,87 @@ private void updateActivated(ModificationEntity entity, boolean activated) {
         }
     }
 
+    @Transactional
+    public void updateRootNetworkApplicability(@NonNull List<UUID> modificationUuids, @NonNull String rootNetworkTag, boolean applicable) {
+        modificationUuids.forEach(modificationUuid -> updateRootNetworkApplicability(getModificationEntity(modificationUuid), rootNetworkTag, applicable));
```
Meklo, line 908: We have a N+1 issue here on select queries by loading modifications one by one, preloading all modifications as a first step would be judicious. Also maybe adding a test to check for query count could be great.

Meklo: The insert/update count may also be affected but I'm not sure

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -41,8 +42,21 @@ public interface ModificationRepository extends JpaRepository<ModificationEntity
...
+    @Query("""
+            SELECT m FROM ModificationEntity m
+            WHERE m.container.id = :containerId AND m.stashed = false AND m.activated = true
+              AND (:rootNetworkTag IS NULL OR NOT EXISTS (
+                  SELECT 1 FROM ModificationEntity holder JOIN holder.applicabilityByRootNetworkTag a
+                  WHERE holder.id = COALESCE((SELECT r.referenceId FROM ModificationReferenceEntity r WHERE r.id = m.id), m.id)
+                    AND KEY(a) = :rootNetworkTag AND VALUE(a) = false))
+            ORDER BY m.modificationsOrder
+            """)
+    List<ModificationEntity> findAllActiveModificationsByContainerId(@Param("containerId") UUID containerId,
```
Meklo, line 58: It may be worth exploring whether this query is well optimized, since it's on the critical path of the node building feature.

I compared the current version against a LEFT JOIN rewrite on a container with 600 modifications, with references and exclusions. Same result (569 rows both ways), but the work done to get there is quite different:

- resolving reference_id through a subquery does one index lookup per candidate modification — 625 lookups here, and it would be ~5000 on a group of 5000. The rewrite does it once.

> SubPlan 1
>   ->  Index Scan using "modification_referencePK" on modification_reference r
>         (actual time=0.000..0.000 rows=0 loops=625)

- the holder join reads the whole modification table just to recover ids we already have. That one grows with the total table rather than the group, so it gets worse over time regardless of group size.

> ->  Index Only Scan using "modificationPK" on modification holder
>       (actual time=0.004..0.257 rows=898 loops=1)

Joining modification_reference directly makes reference_id an ordinary column and both issues go away:

```
SELECT m.*
FROM modification m
LEFT JOIN modification_reference r ON r.id = m.id
LEFT JOIN modification_root_network_applicability a
       ON a.modification_id = COALESCE(r.reference_id, m.id)
      AND a.root_network_tag = :tag
WHERE m.container_id = :containerId
  AND m.stashed = false
  AND m.activated = true
  AND (a.applicable IS NULL OR a.applicable = true)
ORDER BY m.modifications_order;
```

> ->  Hash Left Join  (actual time=0.043..0.191 rows=597 loops=1)
>       Hash Cond: (m.id = r.id)
>       ->  Index Scan using "modification_referencePK" on modification_reference r
>             (actual time=0.004..0.005 rows=2 loops=1)


Timings are both around 1ms at this size so they don't tell us much, the concern is the scaling.
findApplicabilitiesByIdIn has the same shape and is on the modification table loading path, so probably worth applying the same treatment.

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
flomillot: That's what I did first but it is actually useless I think.
https://github.com/gridsuite/network-modification-server/blob/0781ec7e0b47a66bdf567136919a56461f1606d9/src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java#L411
The filter is already done here, no ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
flomillot: Actually this behavior (spreading the applicability to children) is temporary and will be removed in the next ticket, so I don't want to spend too much time to optimize something that will remove in maximum few weeks.

flomillot: However if you were talking only of the originals UIDs then you're right I will load them before

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
flomillot: Thank you for your investigations and suggestions. I change my queries and remove this holder system.

flomillot: I change a bit what you suggest and I use TREAT instead of doing another JOIN because the JOIN on reference columns is already did by Hibernate.

src/main/java/org/gridsuite/modification/server/service/NetworkModificationService.java
Meklo: Yea I missed it no need for additionnal overhead

Meklo (approved):

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
```diff
@@ -41,8 +43,21 @@ public interface ModificationRepository extends JpaRepository<ModificationEntity
...
+     * @return the modifications of the container that are applied on the given root network tag, that is the activated
+     * ones the tag does not deactivate. A reference is resolved to the shared modification carrying the applicability.
+     */
+    @Query("""
+            SELECT m FROM ModificationEntity m
+            LEFT JOIN ModificationRootNetworkApplicabilityEntity a
+                   ON a.modification.id = COALESCE(TREAT(m AS ModificationReferenceEntity).referenceId, m.id)
+                  AND a.rootNetworkTag = :rootNetworkTag
+            WHERE m.container.id = :containerId AND m.stashed = false AND m.activated = true
+              AND COALESCE(a.applicable, true) = true
```
Mathieu-Deharbe: I odn't think that "= true" is needed here.

Meklo (approved):

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -608,6 +683,10 @@ public List<ModificationInfos> getModificationsInfosToExport(List<UUID> groupUui
 
     @Transactional(readOnly = true)
     public ModificationInfos getModificationInfo(UUID modificationUuid) {
+        return addApplicabilities(List.of(toModificationInfo(modificationUuid))).getFirst();
+    }
+
+    private ModificationInfos toModificationInfo(UUID modificationUuid) {
         return toModificationsInfosOptimized(getModificationEntity(modificationUuid));
     }
```
Mathieu-Deharbe: I think this is confusing and a bit dangerous to have sometimes `ModificationInfos` with Applicabilities and sometimes without, with no real way to tell the difference.

Is this a performance issue why you not to put those directly in 
`public ModificationInfos toModificationInfos()`
like `activated` ?

Are there times where the `ModificationInfos` are used with "activated" but without "Applicabilities" ?

Anyway maybe at least precising this in the function name would make it easier to follow because `getModificationInfo` and `toModificationInfo` seem identical from the signature.

Mathieu-Deharbe (commented):
The first example of the ticket, written by Thibault doesn't seem to work anymore :  

1. my applicabilities

Image

2. I delete the "lil" root network : 

Image

3. I add a new root network with the same tag than the deleted one. The applicabilities should be like for the deled one but they are not : 

Image

I am quite sure that this worked during my first test about 1 week ago.

Not that being in a composite or not doesn't change nything.

src/test/java/org/gridsuite/modification/server/service/BuildTest.java
```diff
@@ -1004,6 +1009,91 @@ void runBuildWithReportErrorTest(final MockWebServer server) throws Exception {
...
+     * saved, on the tree as it stands. The composite itself is applicable here: only its content is deactivated.
+     */
+    @Test
+    void applyingACompositeLeavesOutTheContentTheRootNetworkDeactivates(final MockWebServer server) {
+        Network appliedNetwork = NetworkCreation.create(TEST_NETWORK_ID, true);
+        List<ModificationInfos> inserted = insertCompositeCoveringEveryApplicabilityCase();
+
+        applyOnRootNetwork(inserted, TEST_ROOT_NETWORK_TAG, appliedNetwork);
+
+        assertOnlyTheContentTheTagAllowsIsApplied(appliedNetwork);
```
Mathieu-Deharbe: I find the naming weird. Why `appliedNetwork` ?
```suggestion
        Network network = NetworkCreation.create(TEST_NETWORK_ID, true);
        List<ModificationInfos> composites = insertCompositeCoveringEveryApplicabilityCase();

        applyOnRootNetwork(composites, TEST_ROOT_NETWORK_TAG, network);

        assertOnlyTheContentTheTagAllowsIsApplied(appliedNetwork);
```

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
```diff
@@ -404,7 +407,7 @@ void testRestoreNetworkModifications() throws Exception {
                 .equipmentAttributeName("open")
                 .equipmentAttributeValue(true)
                 .equipmentId("v1b1")
-                .stashed(true)
+                .stashed(false)
```
Mathieu-Deharbe: False by default.
```suggestion
```

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
```diff
@@ -482,6 +492,71 @@ void testDisableNetworkModifications() throws Exception {
...
+    void testRenameAndDeleteRootNetworkTag() throws Exception {
+        List<ModificationInfos> modifications = createSomeSwitchModifications(TEST_GROUP_ID, 1);
+        UUID modificationUuid = modifications.getFirst().getUuid();
+        mockMvc.perform(put(URI_NETWORK_MODIF_BASE + "/root-network-applicability")
+                .queryParam("uuids", modificationUuid.toString())
+                .queryParam("rootNetworkTag", ROOT_NETWORK_TAG)
+                .queryParam("applicable", "false")
+        ).andExpect(status().isOk());
+
+        // renaming the root network tag carries along what it deactivates
```
Mathieu-Deharbe:
```suggestion
        // renaming the root network tag carries along its applicabilities
```

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
```diff
@@ -482,6 +492,71 @@ void testDisableNetworkModifications() throws Exception {
...
+        mockMvc.perform(put(URI_NETWORK_MODIF_BASE + "/root-network-tag")
+                .queryParam("groupUuids", TEST_GROUP_ID.toString())
+                .queryParam("oldTag", ROOT_NETWORK_TAG)
+                .queryParam("newTag", RENAMED_ROOT_NETWORK_TAG)
+        ).andExpect(status().isOk());
+        assertEquals(Map.of(RENAMED_ROOT_NETWORK_TAG, false), readApplicabilities(TEST_GROUP_ID).get(modificationUuid));
+
+        // and deleting the root network takes its tag away
```
Mathieu-Deharbe: We don't really "delete the root network" here.

Mathieu-Deharbe (commented):
Review finished. Only minor remarks + the above topic about deleted and recreated root network tags above.

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
flomillot: I agree and yes, it's for performance to avoid too many SQL requests. Check how it's done for composite.

Actually there is no reason to separate them. 

I created a ticket to refactor this part.
But still need to decide if we take the applicabilities every time or if we prefer performance and take it only when we need. 
Like you I prefer the option one but from what I understand the SQL performance is very important and is the priority.

Mathieu-Deharbe (approved):

src/main/java/org/gridsuite/modification/server/repositories/ModificationRepository.java
flomillot: Done

src/test/java/org/gridsuite/modification/server/service/BuildTest.java
flomillot: Yeah, you're right, I changes.

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
flomillot: done

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
flomillot: done

src/test/java/org/gridsuite/modification/server/ModificationControllerTest.java
flomillot: Yes, my bad, done

PR #869, opened 2026-08-13
reference modif actions

souissimai:


Mathieu-Deharbe (changes requested):
Here I imported a composite as shared. Its name is "modif à partager" and then I drag and droped it inside a composite : 

Image

We can see in the database that it is still a STUDY_NODE reference_type. And its reference_id is still the id of its node (I checked) : 

Image

When a modification reference is drag and dropped, its references from directory server should be updated. Here it should become a NETWORK_MODIFICATION and its reference_id should be the id of the composite network modification where it has been dropped.

Mathieu-Deharbe (changes requested):
This one may be hard to handle ! : I have 2 reference modifications in a node : 

Image

Because of this I have two identical references in directory server (see the last 2): 

Image

If I cut one and copy ONE of them to another node and let the other here : both references have been removed from the database, even if one of the 2  references modification is still in the previous node.

A new reference has been added for the pasted reference modification, which is good.

cf maybe : https://github.com/gridsuite/directory-server/pull/263#discussion_r3811480460

src/main/java/org/gridsuite/modification/server/dto/ReferenceData.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.UUID;
+
+/**
+ * One occurrence of a network modification referencing a shared composite modification.
+ *
+ * @param modificationUuid uuid of the modification-reference itself
+ * @param referenceId uuid of the referenced shared composite in directory-server
+ * @param containerId uuid of the composite containing the reference, null if the modification-reference is at the root level
+ * @author RTE
```
Mathieu-Deharbe: Name yourself !

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -1082,6 +1081,20 @@ private AbstractModificationContainerEntity getContainer(ModificationContainerIn
...
+     */
+    private UUID resolveContainerId(ModificationContainerInfos containerInfos) {
+        if (ModificationContainerType.COMPOSITE.equals(containerInfos.type())) {
+            return modificationRepository.findById(containerInfos.id())
+                    .filter(ModificationReferenceEntity.class::isInstance)
+                    .map(entity -> ((ModificationReferenceEntity) entity).getReferenceId())
+                    .orElse(containerInfos.id());
+        }
+        return containerInfos.id();
+    }
```
Mathieu-Deharbe, line 1096: I don't think that this function is useful. 

`containerInfos.id()` should never be a reference modification uuid, it should be the uuid of the real container, that is the uuid of the composite toward which the reference is pointing. 

Therefore `.filter(ModificationReferenceEntity.class::isInstance)` should always return false. Or Something strange is in the container table.

Mathieu-Deharbe: Ok I misunderstood : `resolveContainerId` allows to get the composite uuid container from the reference modification uuid **before** geting it from the container repository : 

`AbstractModificationContainerEntity containerEntity = modificationContainerRepository.findById(containerId)`

So the reference is never a container in the `modificationContainerRepository` but it can be one in the dto.

This is a bit hard to follow. Maybe we could use different terms, but given that this is going to be refactored, we may improve this later.

Mathieu-Deharbe: Just to make it even clearer : 
```suggestion
    /**
     * A COMPOSITE container may be designated through a modification reference, in the ModificationContainerInfos dto: 
     * in that case the actual container is the shared composite the reference points to.
     */
    private UUID resolveContainerId(ModificationContainerInfos containerInfos) {
        if (ModificationContainerType.COMPOSITE.equals(containerInfos.type())) {
            return modificationRepository.findById(containerInfos.id())
                    .filter(ModificationReferenceEntity.class::isInstance)
                    .map(entity -> ((ModificationReferenceEntity) entity).getReferenceId())
                    .orElse(containerInfos.id());
        }
        return containerInfos.id();
    }
```
But I really think that we should add a "TODO" about this because we are forcing one more request for all the composites when it is not really needed : the container may be typed as REFERENCE in the dto, or the actual shared composite container may be determined in the front if we add the data in the front...

souissimai: right, this should be simplified later

src/main/java/org/gridsuite/modification/server/dto/ReferenceData.java
souissimai: ok

Mathieu-Deharbe (approved):

PR #871, opened 2026-08-18
feat(serde-network-modification): add GET endpoint for abstract network modifications

KoloMenek:


src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -147,6 +148,17 @@ public ResponseEntity<Void> deleteModificationGroup(@Parameter(description = "Gr
         return ResponseEntity.ok().build();
     }
 
+    @GetMapping(value = "/abstract-network-modification/{uuid}", produces = MediaType.APPLICATION_JSON_VALUE)
```
TheMaskedTurtle: Maybe for endpoints structure, something like `/network-modifications/{uuid}/standalone` would be clearer

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -147,6 +148,17 @@ public ResponseEntity<Void> deleteModificationGroup(@Parameter(description = "Gr
...
+    @Operation(summary = "Get an abstract network modification")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = "The abstract network modification was returned"),
+        @ApiResponse(responseCode = "404", description = "The abstract network modification was not found")
+    })
```
TheMaskedTurtle: Maybe also replace abstract by standalone. Abstract is not a reliable user information, the final point of this is that the DTO is standalone

TheMaskedTurtle (changes requested):

src/main/java/org/gridsuite/modification/server/NetworkModificationController.java
```diff
@@ -147,6 +148,30 @@ public ResponseEntity<Void> deleteModificationGroup(@Parameter(description = "Gr
...
+    }
+
+    @GetMapping(value = "/network-modifications/standalone", produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(summary = "Get an abstract network modification")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = "The standalone network modification was returned"),
+        @ApiResponse(responseCode = "404", description = "The standalone network modification was not found")
+    })
+    public ResponseEntity<Map<UUID, AbstractModification>> getAbstractNetworkModifications(
+        @Parameter(description = "Network modification UUIDs") @RequestParam("uuids") List<UUID> networkModificationUuids,
```
TheMaskedTurtle, line 171: This would be limited by size, in some other cases we did that we converting into a POST with the ID list as request body.

TheMaskedTurtle (approved):

PR #874, opened 2026-08-19
Add an endpoint sharing a composite modification of a group

flomillot:
Adds `POST /v1/network-composite-modifications/{uuid}/share?groupUuid=&name=`, used when a composite modification of a study node becomes a shared modification stored in GridExplore.

In a single transaction, the composite modification is taken out of its group and a `MODIFICATION_REFERENCE` to it takes its place, at the very same order. The composite keeps its own uuid — it is moved, not duplicated — so the caller can store it in the directory server under that uuid, and it is renamed when a name is given.

Sharing a modification which is not a composite one is rejected with the new `MODIFICATION_BAD_TYPE` business error, and sharing one which is not owned by the given group with `MODIFICATION_NOT_FOUND`.

### DIFF
--- a/src/main/java/org/gridsuite/modification/server/CompositeController.java
+++ b/src/main/java/org/gridsuite/modification/server/CompositeController.java
@@ -84,12 +84,13 @@ public class CompositeController {
 
     @PostMapping(value = "/{uuid}/share")
     @Operation(summary = "Extract a composite modification from its group, replacing it by a reference to it")
-    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The composite modification has been extracted, its uuid is returned")})
-    public ResponseEntity<UUID> extractCompositeModificationToShare(
+    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The composite modification has been extracted")})
+    public ResponseEntity<Void> extractCompositeModificationToShare(
             @PathVariable("uuid") UUID compositeModificationUuid,
             @Parameter(description = "Group owning the composite modification", required = true) @RequestParam("groupUuid") UUID groupUuid,
             @Parameter(description = "New name of the shared composite modification") @RequestParam(value = "name", required = false) String name) {
-        return ResponseEntity.ok().body(networkModificationService.extractCompositeModificationToShare(groupUuid, compositeModificationUuid, name));
+        networkModificationService.extractCompositeModificationToShare(groupUuid, compositeModificationUuid, name);
+        return ResponseEntity.ok().build();
     }
 
     @GetMapping(value = "/network-modifications", produces = MediaType.APPLICATION_JSON_VALUE)
--- a/src/main/java/org/gridsuite/modification/server/CompositeController.java
+++ b/src/main/java/org/gridsuite/modification/server/CompositeController.java
@@ -82,6 +82,16 @@ public class CompositeController {
         return ResponseEntity.ok().body(networkModificationService.createNetworkCompositeModification(modificationUuids, name));
     }
 
+    @PostMapping(value = "/{uuid}/share")
+    @Operation(summary = "Extract a composite modification from its group, replacing it by a reference to it")
+    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The composite modification has been extracted, its uuid is returned")})
+    public ResponseEntity<UUID> extractCompositeModificationToShare(
+            @PathVariable("uuid") UUID compositeModificationUuid,
+            @Parameter(description = "Group owning the composite modification", required = true) @RequestParam("groupUuid") UUID groupUuid,
+            @Parameter(description = "New name of the shared composite modification") @RequestParam(value = "name", required = false) String name) {
+        return ResponseEntity.ok().body(networkModificationService.extractCompositeModificationToShare(groupUuid, compositeModificationUuid, name));
+    }
+
     @GetMapping(value = "/network-modifications", produces = MediaType.APPLICATION_JSON_VALUE)
     @Operation(summary = "Get the list of all the network modifications inside a list of composite modifications")
     @ApiResponse(responseCode = "200", description = "Map of modifications inside the composite modifications for each composite")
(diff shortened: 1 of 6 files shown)


src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -1043,6 +1044,51 @@ public List<ModificationInfos> insertCompositeModifications(
...
+        if (!(modificationEntity instanceof CompositeModificationEntity compositeEntity)) {
+            String expectedType = ModificationType.COMPOSITE_MODIFICATION.name();
+            throw new NetworkModificationServerException(MODIFICATION_BAD_TYPE,
+                String.format(MODIFICATION_BAD_TYPE.messageTemplate(), modificationUuid, modificationEntity.getType(), expectedType),
+                Map.of(MODIFICATION_ID, modificationUuid.toString(), "modificationType", modificationEntity.getType(), "expectedModificationType", expectedType));
+        }
+        if (!groupUuid.equals(modificationEntity.getContainerUuid())) {
+            throw new NetworkModificationServerException(MODIFICATION_NOT_FOUND,
+                String.format("Modification %s is not owned by group %s", modificationUuid, groupUuid),
+                Map.of(MODIFICATION_ID, modificationUuid + " (group = " + groupUuid + ")"));
```
ghazwarhili: Map.of(MODIFICATION_ID, modificationUuid, "groupId", groupUuid)) ?
(Jackson serializes UUID as a string anyway)

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -1043,6 +1044,51 @@ public List<ModificationInfos> insertCompositeModifications(
...
+        ModificationEntity modificationEntity = getModificationEntity(modificationUuid);
+        if (!(modificationEntity instanceof CompositeModificationEntity compositeEntity)) {
+            String expectedType = ModificationType.COMPOSITE_MODIFICATION.name();
+            throw new NetworkModificationServerException(MODIFICATION_BAD_TYPE,
+                String.format(MODIFICATION_BAD_TYPE.messageTemplate(), modificationUuid, modificationEntity.getType(), expectedType),
+                Map.of(MODIFICATION_ID, modificationUuid.toString(), "modificationType", modificationEntity.getType(), "expectedModificationType", expectedType));
+        }
+        if (!groupUuid.equals(modificationEntity.getContainerUuid())) {
+            throw new NetworkModificationServerException(MODIFICATION_NOT_FOUND,
+                String.format("Modification %s is not owned by group %s", modificationUuid, groupUuid),
```
ghazwarhili, line 1067: String.format(MODIFICATION_NOT_FOUND.messageTemplate(), modificationUuid) ?

src/main/java/org/gridsuite/modification/server/CompositeController.java
```diff
@@ -82,6 +82,16 @@ public ResponseEntity<UUID> createNetworkCompositeModification(@Parameter(descri
         return ResponseEntity.ok().body(networkModificationService.createNetworkCompositeModification(modificationUuids, name));
     }
 
+    @PostMapping(value = "/{uuid}/share")
+    @Operation(summary = "Extract a composite modification from its group, replacing it by a reference to it")
```
ghazwarhili, line 86: Extract the composite modification from the group and replace it with a reference ?

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
```diff
@@ -1043,6 +1044,51 @@ public List<ModificationInfos> insertCompositeModifications(
...
+        groupEntity.addModification(referenceEntity, compositeEntity.getModificationsOrder());
+        groupEntity.removeModifications(List.of(modificationUuid));
+        compositeEntity.setContainer(null);
+        compositeEntity.setModificationsOrder(0);
+        if (name != null) {
+            compositeModificationRepository.renameCompositeModification(compositeEntity, name);
+        }
+
+        modificationRepository.save(referenceEntity);
+        modificationRepository.save(compositeEntity);
```
ghazwarhili, line 1088: modificationRepository.save(referenceEntity);
modificationRepository.save(compositeEntity);

are you sure about those save ? they seem a little weird to me do you really need to perform the save operation after addModification? (the modifications are already committed)

ghazwarhili (changes requested):

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
flomillot: Took same as existing
https://github.com/gridsuite/network-modification-server/blob/37d96b3e26b10c4a35515a5a80503402b26d415a/src/main/java/org/gridsuite/modification/server/entities/AbstractModificationContainerEntity.java#L85

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
flomillot: Nope it's a different one.

src/main/java/org/gridsuite/modification/server/CompositeController.java
flomillot: Honestly I don't see the difference and I prefer the original one

src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java
flomillot: It's a good catch and actually technically it works without. However, it totally depends on how other entities are annotated. 
https://github.com/gridsuite/network-modification-server/blob/37d96b3e26b10c4a35515a5a80503402b26d415a/src/main/java/org/gridsuite/modification/server/entities/AbstractModificationContainerEntity.java#L38
It's tricky for me and keeping the save is safer even if it's useless right now.

Moreover, there is already a similar case keeping the save. 
https://github.com/gridsuite/network-modification-server/blob/183c3e9273f7dedcea0df94ba8870048a218b571/src/main/java/org/gridsuite/modification/server/repositories/NetworkModificationRepository.java#L1172

Maybe we should ask a tech lead about it.

ghazwarhili: @etiennehomer what do you think ?

flomillot: Lines are removed

ghazwarhili (approved):

